PR #286, opened 2025-03-12
Require RBC votes to carry evidence.

twittner:
Like with vertex proposals we want to prevent dishonest parties to send us votes arbitrarily many rounds ahead, which we then need to store and process. Therefore we require every vote to come with evidence taken from the message this vote applies to, so we can be sure a quorum of parties is behind the previous round. To avoid excessive signature checks we only verify the evidence signatures for rounds where we do not already have some data, as the evidence plays no role in RBC voting.

### DIFF
--- a/sailfish-rbc/src/abraham/worker.rs
+++ b/sailfish-rbc/src/abraham/worker.rs
@@ -9,7 +9,7 @@ use bytes::{BufMut, Bytes, BytesMut};
 use committable::{Commitment, Committable};
 use multisig::{Certificate, Envelope, PublicKey, VoteAccumulator};
 use multisig::{Unchecked, Validated};
-use sailfish_types::{Evidence, Message, MessageKind, RawComm, RoundNumber, Vertex};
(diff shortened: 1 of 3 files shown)


sailfish-rbc/src/abraham/worker.rs
```diff
@@ -641,6 +651,19 @@ impl<C: RawComm, T: Clone + Committable + Serialize + DeserializeOwned> Worker<C
         let commit = digest.commit();
         let source = *env.signing_key();
 
+        // For unknown rounds we check the vote evidence to make sure
+        // a quorum of parties is backing the round prior to the vote.
+        if !self.buffer.contains_key(&digest.round()) {
```
akonring, line 656: thought: IIUC the verification immediately kicks in if the digest is the first message in every round. Would it be appropriate to relax this requirement such that we only verify if the gap is > 1 round.

akonring (approved):
Nice.

PR #291, opened 2025-03-14
Selectively re-send messages and votes.

twittner:
Instead of re-broadcasting the message or our vote when revisiting the status of messages, find out which voters are outstanding and send the message or vote only to them. Spamming those who can only discard the data serves no purpose.

Also exercise back-pressure on the network. Previously, if the capacity of the channel to the application layer drops to 0 the worker waits. If the application layer is only consuming from that channel after it has sent, this can result in a deadlock. Therefore we now only accept inbound messages from the network if the channel capacity is greater than 0.


akonring (approved):
LGTM.

PR #294, opened 2025-03-20
Add transaction order test.

twittner:
Small integration test to check that the sequencer committee produces the same sequence of transactions.

### DIFF
--- a/timeboost-sequencer/src/include.rs
+++ b/timeboost-sequencer/src/include.rs
@@ -1,6 +1,6 @@
 use std::cmp::max;
 use std::collections::btree_map::Entry;
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, HashSet};
 
 use multisig::Committee;
 use sailfish::types::RoundNumber;
@@ -68,7 +68,7 @@ impl Includer {
             self.seqno = SeqNo::zero();
         }
 
-        let mut transactions: HashMap<Transaction, usize> = HashMap::new();
+        let mut transactions: BTreeMap<Transaction, usize> = BTreeMap::new();
         let mut bundles: BTreeMap<SeqNo, PriorityBundle> = BTreeMap::new();
         let mut retry = RetryList::new();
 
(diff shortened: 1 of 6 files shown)


timeboost-sequencer/src/include.rs
```diff
@@ -68,7 +68,7 @@ impl Includer {
             self.seqno = SeqNo::zero();
         }
 
-        let mut transactions: HashMap<Transaction, usize> = HashMap::new();
+        let mut transactions: BTreeMap<Transaction, usize> = BTreeMap::new();
```
twittner, line 71: While non-priority transactions are defined as an unordered set, an ordering is required here as an implementation detail. The digest of an inclusion list is used as the seed in the ordering phase, so parties need to be able to calculate the same digest.

timeboost-sequencer/src/lib.rs
```diff
@@ -218,27 +218,44 @@ impl Task {
             select! {
                 result = self.sailfish.next() => match result {
                     Ok(actions) => {
-                        let mut payloads: BTreeMap<RoundNumber, Vec<CandidateList>> = BTreeMap::new();
```
twittner, line 221: As soon as a quorum of parties has reached consensus, they will produce deliver actions. However the candidate lists of late parties may appear in later iterations. In order to guarantee strictly monotonic round numbers all consecutive deliver actions need to be passed to the inclusion phase, together with their highest round number. This will put the late arrivals together with the next round quorum candidates. The previous implementation would have passed the late candidate list(s) alone to the includer, potentially producing an inclusion list with the same round number.

akonring (approved):

timeboost-sequencer/src/include.rs
akonring: (apologize; the above is an old comment we already discussed briefly but is still relevant imo)

twittner: The seed has yet to be defined. Using the hash of the inclusion list was an interim solution which might very well change.

PR #296, opened 2025-03-26
Timeboost Bundle Format

akonring:
This PR is an implementation of the Bundle-centric approach for Timeboost transactions and roughly follows the design proposal (see [Design Proposal](https://docs.google.com/document/d/1Ngos6x090UEDdugeprmydzaSAqX_W0MlBMmuZ-pZkfU/edit?usp=sharing) for detailed description).

### tl;dr:
1. No need to "shoe-horn" bundles and encrypted transactions into Ethereum transaction format.
2. While the spec addendum describes a union type (priority bundles, encrypted non-priority tx, regular non-priority tx), this PR (and design proposal) instead introduces non-priority bundles.
3. Priority bundle submission should closely mirror centralized timeboost; the idea being that a PLC should be able to use the same approach to submit a priority bundle to the new decentralized timeboost.

### Content
This PR refactors timeboost phases to handle the new format for  (regular) `Bundle` and modified `PriorityBundle`. Most of the underlying logic remains unchanged.

### Other points
- This PR touches almost all parts of the Timeboost stack and thus will not be easy to maintain. 
- Final confirmation is needed for this design to be merged. In any case, modifying the new design to conform to the union type (described above) is a trivial addition to this PR.


timeboost-sequencer/Cargo.toml
```diff
@@ -6,6 +6,10 @@ edition.workspace = true
 rust-version.workspace = true
 
 [dependencies]
+alloy = { workspace = true }
+alloy-rlp = { workspace = true }
+alloy-signer = { workspace = true }
+alloy-signer-local = { workspace = true }
```
twittner, line 12:
```suggestion
alloy-consensus = { workspace = true }
```

timeboost-types/Cargo.toml
```diff
@@ -5,13 +5,14 @@ version.workspace = true
 edition.workspace = true
 rust-version.workspace = true
 
-[features]
-arbitrary = ["dep:arbitrary", "alloy-primitives/arbitrary"]
-
 [dependencies]
-alloy-primitives = { workspace = true }
+alloy = { workspace = true }
```
twittner, line 9:
```toml
alloy-consensus = { workspace = true }
alloy-primitives = { workspace = true }
alloy-rlp = { workspace = true }
alloy-signer = { workspace = true }
alloy-signer-local = { workspace = true }
```

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,463 @@
...
+const DOMAIN: &str = "TIMEBOOST_BID";
+
+#[derive(Debug, Clone)]
+pub enum BundleVariant {
+    Regular(Bundle),
+    Priority(PriorityBundle<Signed>),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Bundle {
```
twittner, line 21: Perhaps this should be named `RegularBundle` and `BundleVariant` can become `Bundle`?

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,463 @@
...
+        RawCommitmentBuilder::new("PriorityBundle")
+            .field("chain", self.chain_id().commit())
+            .field("epoch", self.epoch().commit())
+            .var_size_field("data", self.data())
+            .field("keysetid", self.kid().unwrap_or_default().commit())
+            .finalize()
+    }
+}
+
+pub trait State {}
```
twittner, line 112: What is the purpose of this trait? AFAICT it is sufficient to have `Signed` and `Unsigned` type definitions.

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,463 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Signed {
+    signature: Signature,
+}
+pub struct Unsigned {}
```
twittner, line 117:
```suggestion
pub struct Unsigned(());
```

Maybe the private field can be added to ensure that `Unsigned` values can only be constructed in here?

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,463 @@
...
+impl State for Signed {}
+impl State for Unsigned {}
```
twittner, line 119: I do not think the `State` trait is necessary.

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,463 @@
...
+    pub fn auction(&self) -> Address {
+        self.auction
+    }
+
+    pub fn seqno(&self) -> SeqNo {
+        self.seqno
+    }
+
+    // https://github.com/OffchainLabs/nitro/blob/1e16dc408d24a7784f19acd1e76a71daac528a22/timeboost/types.go#L206
+    pub fn as_bytes(&self) -> Vec<u8> {
```
twittner, line 144:
```suggestion
    pub fn to_bytes(&self) -> Vec<u8> {
```
https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

timeboost-types/src/time.rs
```diff
@@ -17,6 +17,10 @@ impl Epoch {
     pub fn now() -> Self {
         Timestamp::now().epoch()
     }
+
+    pub fn as_bytes(&self) -> [u8; 8] {
+        self.0.to_be_bytes()
+    }
```
twittner, line 23:
```suggestion
```

timeboost-types/src/seqno.rs
```diff
@@ -18,6 +18,10 @@ impl SeqNo {
     pub fn is_zero(self) -> bool {
         self.0 == 0
     }
+
+    pub fn as_bytes(&self) -> [u8; 8] {
+        self.0.to_be_bytes()
+    }
```
twittner, line 24:
```suggestion
```

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,463 @@
...
+        let mut buf = Vec::new();
+        buf.extend_from_slice(DOMAIN.as_bytes());
+        buf.extend_from_slice(&self.bundle().chain.0.to_be_bytes());
+        buf.extend_from_slice(self.auction.0.0.as_slice());
+        buf.extend_from_slice(&self.bundle().epoch.as_bytes());
+        buf.extend_from_slice(&self.seqno.as_bytes());
```
twittner, line 150:
```suggestion
        buf.extend_from_slice(&self.bundle().epoch.to_be_bytes());
        buf.extend_from_slice(&self.seqno.to_be_bytes());
```

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,463 @@
...
+        self.hash = digest.into();
+    }
+
+    #[cfg(feature = "arbitrary")]
+    pub fn arbitrary(
+        u: &mut arbitrary::Unstructured<'_>,
+    ) -> arbitrary::Result<PriorityBundle<Signed>> {
+        let bundle = Bundle::arbitrary(u)?;
+        let auction = Address::default();
+        let seqno = SeqNo::from(u.int_in_range(1..=u64::MAX)?);
```
twittner, line 252: If the sequence number space is that large, chances are that there will always be gaps in tests.

timeboost-sequencer/src/lib.rs
```diff
@@ -170,7 +170,7 @@ impl Sequencer {
 
     pub fn add_transactions<I>(&mut self, it: I)
```
twittner, line 171:
```suggestion
    pub fn add_bundles<I>(&mut self, it: I)
```

timeboost-sequencer/src/queue.rs
```diff
@@ -57,50 +57,50 @@ impl TransactionQueue {
...
+                // TODO: Check auction contract address on bundle
+                {
+                    match b.validate(epoch_now, Some(inner.priority_addr)) {
+                        Ok(_) => {
+                            let epoch = b.bundle().epoch();
+                            inner.priority.entry(epoch).or_default().push(b);
+                        }
+                        Err(e) => {
+                            trace!(
+                                "Validation failed for bundle signed by: {:?}, error: {:?}",
```
twittner, line 84: No need to do string interpolation. `trace!` accepts structured parameters, e.g.

```rust
trace!(signer = ?b.sender(), err = %e, "bundle validation failed")
```

twittner (commented):
Thanks for putting so much effort into getting the bundle format details right!

Cargo.toml
```diff
@@ -28,8 +28,10 @@ opt-level = 3
 
 [workspace.dependencies]
 aes-gcm = { version = "0.10.3" }
-alloy-primitives = { version = "0.8.22", features = ["serde", "rlp"] }
+alloy = { version = "0.12.6", features = ["arbitrary", "rlp", "k256"] }
```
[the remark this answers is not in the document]
akonring: good point. fixed.

timeboost-types/src/bundle.rs
akonring: This was merely to control the type state without introducing much duplicate impl-code. Yes, introducing State was not strictly necessary but (imo) provided good readability. I've modified the approach to use standard type composition while preserving the type state pattern.

timeboost-types/src/bundle.rs
akonring: Thanks for the pointer. Fixed.

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,467 @@
...
+        };
+
+        let sender = self.sender()?;
+        if sender.0 != plc.0 {
+            return Err(ValidationError::NotPLC);
+        }
+        Ok(())
+    }
+
+    pub fn sender(&self) -> Result<Address, ValidationError> {
```
jparr721, line 224: Can we get a doc comment here? It's not super clear why this is a failible operation.

timeboost-types/src/bundle.rs
jparr721: Addendum: do we hit any tricky behavior when we hit u64::MAX in anything? It's overwhelmingly unlikely that we'll hit this, of course, but just a thought. Not sure if we do any math anywhere.

timeboost/api/endpoints.toml
```diff
@@ -3,10 +3,15 @@ NAME = "timeboost submit api"
...
 
-[route.submit]
-PATH = ["/submit"]
+[route.submit-priority]
+PATH = ["/submit-priority"]
 METHOD = "POST"
-DOC = "Submit transaction to Timeboost"
+DOC = "Submit priority bundle to Timeboost"
+
+[route.submit-regular]
```
jparr721, line 11: Is this correct? Would a user ever submit a priority bundle to the node? Doesn't the user just submit a transaction and *only* the PLC has priority tx? How do we not spoof this?

jparr721 (commented):
Some comments

timeboost-types/src/bundle.rs
akonring: > If the sequence number space is that large, chances are that there will always be gaps in tests.

Fixed: [24ec63e](https://github.com/EspressoSystems/timeboost/pull/296/commits/24ec63e16bd42217bfef94f2b1688f94c050b390).
But there will be gaps, regardless. We should fix test/validation - currently no priority bundles make it through the inclusion phase. But consider this out-of-scope for this PR.

> Addendum: do we hit any tricky behavior when we hit u64::MAX in anything? It's overwhelmingly unlikely that we'll hit this, of course, but just a thought. Not sure if we do any math anywhere.

iiuc it is the responsibility of the PLC to ensure correct bundle format (incl. correct seqno). We could do saturating/check add in places like: https://github.com/EspressoSystems/timeboost/blob/24ec63e16bd42217bfef94f2b1688f94c050b390/timeboost-sequencer/src/include.rs#L172 Think Rust does silent overflows by default in release-mode which also seems ok ( don't see how an adversary should exploit this).

timeboost-sequencer/src/include.rs
```diff
@@ -21,7 +24,7 @@ pub struct Includer {
     /// Consensus delayed inbox index.
     index: DelayedInboxIndex,
     /// Cache of transaction hashes for the previous 8 rounds.
-    cache: BTreeMap<RoundNumber, HashSet<Hash>>,
+    cache: BTreeMap<RoundNumber, HashSet<[u8; 32]>>,
```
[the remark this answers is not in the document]
akonring: This is just a 256-bit keccak hash.

timeboost-sequencer/src/decrypt.rs
```diff
@@ -158,49 +165,46 @@ fn assemble_incl(
...
 
-    let mut new_txs = BTreeSet::from_iter(txs);
-    for (i, tx) in encrypted_txs.into_iter().enumerate() {
-        let mut decrypted_tx = new_txs.take(&tx).ok_or(DecryptError::InvalidMessage)?;
+    let mut new_bundles = BTreeSet::from_iter(regular_bundles);
+    for (i, tx) in encrypted_bundles.into_iter().enumerate() {
+        let mut decrypted_bundles = new_bundles.take(&tx).ok_or(DecryptError::InvalidMessage)?;
         let data = dec
-            .get(i + ptxs.len())
+            .get(i + priority_bundles.len())
```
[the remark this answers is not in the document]
akonring: added a comment to the function to make it easier to parse.

timeboost-sequencer/src/queue.rs
```diff
@@ -57,50 +57,46 @@ impl TransactionQueue {
...
-            let epoch = t.nonce().to_epoch();
-
-            if epoch < inner.time.epoch() || epoch > inner.time.epoch() + 1 {
-                continue;
+        let epoch_now = inner.time.epoch();
+        for b in it.into_iter() {
+            match b {
+                BundleVariant::Regular(b) => inner.regular.push_back((now, b)),
+                BundleVariant::Priority(b) =>
+                // TODO: Check auction contract address on bundle
```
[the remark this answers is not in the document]
akonring: I'll remove it from here since the TODO is already present in `Bundle::validate`.
The check in centralized timeboost: https://github.com/OffchainLabs/nitro/blob/1e16dc408d24a7784f19acd1e76a71daac528a22/execution/gethexec/express_lane_service.go#L414.
Applying Occam's razor it seems like an extra check in case of multiple auctions on the same chain.

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,467 @@
...
+            hash: [0; 32],
+            bundle,
+        }
+    }
+
+    pub fn bundle(&self) -> &Bundle {
+        &self.bundle
+    }
+
+    pub fn auction(&self) -> Address {
```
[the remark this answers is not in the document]
akonring: sure. rename sounds good.

timeboost-types/src/bundle.rs
akonring: fixed.

timeboost/api/endpoints.toml
akonring: How do you envision a PLC submitting priority bundles if not through an API endpoint? The bundle will eventually be authenticated (PLC signature will be checked) before being queued.

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub fn digest(&self) -> &[u8; 32] {
+        &self.hash
+    }
+
+    pub fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    // https://github.com/OffchainLabs/nitro/blob/1e16dc408d24a7784f19acd1e76a71daac528a22/execution/gethexec/express_lane_service.go#L309
+    pub fn validate(&self, epoch: Epoch, plc: Option<Address>) -> Result<(), ValidationError> {
```
twittner, line 199: I wonder if `validate` should be a method of `Bundle` or if `Bundle`s can be validated in multiple ways (potentially).

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,470 @@
...
+        this
+    }
+
+    fn update_hash(&mut self) {
+        let digest = self.commit();
+        self.hash = digest.into();
+    }
+
+    #[cfg(feature = "arbitrary")]
+    pub fn arbitrary(
```
twittner, line 254: Again, I think an `Arbitrary` impl would be preferrable.

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,467 @@
...
+use alloy_rlp::Decodable;
+use alloy_signer::{Error, SignerSync, k256::ecdsa::SigningKey};
+use alloy_signer_local::PrivateKeySigner;
+use committable::{Commitment, Committable, RawCommitmentBuilder};
+use serde::{Deserialize, Serialize};
+use timeboost_crypto::KeysetId;
+
+use crate::{Bytes, Epoch, SeqNo};
+
+const DOMAIN: &str = "TIMEBOOST_BID";
```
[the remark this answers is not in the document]
akonring: Domain value as in centralized timeboost:
https://github.com/OffchainLabs/nitro/blob/59659b246d4b79a13c3d82df5b676ce83554bef5/timeboost/auctioneer.go#L33

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,470 @@
...
+        &self.hash
+    }
+
+    pub fn set_data(&mut self, d: Bytes) {
+        self.data = d;
+        self.update_hash()
+    }
+
+    #[cfg(feature = "arbitrary")]
+    pub fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Bundle> {
```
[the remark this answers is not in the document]
akonring: Yes, thank you. Fixed.

timeboost-types/src/bundle.rs
akonring: Not sure what you mean here. Only PriorityBundles need this kind of validation I think. 
Maybe we can adjust and extend validation in future PRs when we have more clarity.

twittner: Previously the validation happened elsewhere by inspecting bundle data. I was not sure why it was moved here into a method when validation does not require private access and there can potentially be multiple ways (abstractly speaking) to validate a value of this type. Just my 2 cents and not important.

timeboost-types/src/bundle.rs
```diff
@@ -0,0 +1,472 @@
+use std::ops::{Deref, DerefMut};
+
+use alloy_consensus::TxEnvelope;
+
+use alloy_rlp::Decodable;
+use alloy_signer::{Error, SignerSync, k256::ecdsa::SigningKey};
+use alloy_signer_local::PrivateKeySigner;
+use arbitrary::{Arbitrary, Result, Unstructured};
```
twittner, line 8:
```suggestion
#[cfg(feature = "arbitrary")]
use arbitrary::{Arbitrary, Result, Unstructured};
```

timeboost-types/Cargo.toml
```diff
@@ -5,6 +5,9 @@ version.workspace = true
 edition.workspace = true
 rust-version.workspace = true
 
+[features]
+default = ["arbitrary"]
```
twittner, line 9: Rather than making this test-related feature a default I think the relevant code sections should be feature-gated.

akonring: Thanks. That is indeed better. Fixed in https://github.com/EspressoSystems/timeboost/pull/296/commits/e855b6f768ea5e3cafde0a863b266ec499773e4a.

twittner (approved):
:+1:

PR #298, opened 2025-03-27
Run demo with encrypted bundles

akonring:
### Content
This PR extends the current load generator so it also encrypts a subset of (non-)priority bundles.

### Additional Changes
- Encryption scheme has been updated such that encryption (`encrypt`) only takes the keysetId as argument and not the full Keyset definition (committee size is not relevant for encryption).
- Introduce a BTreeMap into the Incubator to avoid having duplicate shares when combining.
- Updates local.json to ensure correct decryption for committees of size 5 (Note that decryption will not work for other committee sizes until we have adjusted our config/keyset approach).
- Fixes a bug in our decryption threshold to correctly compute ceiling division.

### DIFF
--- a/timeboost-sequencer/src/lib.rs
+++ b/timeboost-sequencer/src/lib.rs
@@ -23,7 +23,7 @@ use tracing::{Level, error, trace};
 
 use decrypt::{DecryptError, Decrypter};
 use include::Includer;
-use queue::TransactionQueue;
+use queue::BundleQueue;
 use sort::Sorter;
 
 type Result<T> = std::result::Result<T, TimeboostError>;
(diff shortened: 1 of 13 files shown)


timeboost-sequencer/src/lib.rs
```diff
@@ -88,7 +88,7 @@ impl Drop for Sequencer {
 
 struct Task {
     label: PublicKey,
-    transactions: TransactionQueue,
+    transactions: BundleQueue,
```
twittner, line 91:
```suggestion
    bundles: BundleQueue,
```
?

twittner (approved):
:+1:

PR #300, opened 2025-03-28
Track connect tasks and improve log outputs.

twittner:


cliquenet/src/lib.rs
```diff
@@ -164,6 +172,19 @@ struct Server<T: tcp::Listener> {
     ping_interval: Interval,
 }
 
+/// A connect task.
+#[derive(Debug)]
+struct ConnectTask {
+    h: AbortHandle,
```
jparr721, line 178: Perhaps a slightly longer name? :)

cliquenet/src/lib.rs
```diff
@@ -258,6 +276,9 @@ impl Network {
             peers.insert(k, a.into());
         }
 
+        let (otx, orx) = mpsc::channel(PEER_CAPACITY * peers.len());
```
jparr721, line 279: What's the logic behind this calculation? Is there any big difference between this and 10k? I recognize that it's a scaled factor of the peer size, but at the deployment sizes we expect, the computed value is nearly always << 10,000. Just curious.

cliquenet/src/lib.rs
```diff
@@ -286,6 +309,10 @@ impl Network {
 
     /// Send a message to a party, identified by the given public key.
     pub async fn unicast(&self, to: PublicKey, msg: Bytes) -> Result<()> {
+        if msg.len() > MAX_TOTAL_SIZE {
+            warn!(node = %self.label, %to, len = %msg.len(), "message too large to send");
```
jparr721, line 313: Does it matter if this happens? How could we encounter this? Not implying you're doing something wrong, just checking that the log here is not actually potentially a more-severe alert.

cliquenet/src/lib.rs
```diff
@@ -391,22 +422,27 @@ where
...
+                                self.task2key.remove(&id);
+                                continue
+                            };
+                            self.on_connect_task_end(id);
+                            // We only keep the connection if our key is larger than the remote,
+                            // or if we do not have a connection for that key at the moment.
+                            if k < self.key || !self.active.contains_key(&k) {
+                                self.spawn_io(k, s, t)
+                            } else {
+                                warn!(node = %self.key, peer = %k, "dropping new connection");
```
jparr721, line 438: Is this a warning? Or is it benign when this happens?

cliquenet/src/lib.rs
```diff
@@ -391,22 +422,27 @@ where
...
+                            }
                         }
-                    }
-                    Err(e) => {
-                        if !e.is_cancelled() {
-                            error!(node = %self.key, err = %e, "connect task panic");
+                        Err(e) => {
+                            if !e.is_cancelled() {
+                                error!(node = %self.key, err = %e, "connect task panic");
```
jparr721, line 443: Is this un-recoverable? What happens?

cliquenet/src/lib.rs
```diff
@@ -509,14 +584,20 @@ where
...
         let x = self.index.get_by_left(&k).expect("known public key");
         let a = self.peers.get(&k).expect("known address");
-        self.connect_tasks.spawn(connect(
+        let h = self.connect_tasks.spawn(connect(
             (self.key, self.keypair.clone()),
             (k, *x),
             a.clone(),
             self.metrics.clone(),
         ));
+        assert!(self.task2key.insert(h.id(), k).is_none());
```
jparr721, line 599: Do we want to do this? Should we perhaps also log something descriptive if we do die?

sailfish-rbc/src/abraham/worker.rs
```diff
@@ -950,15 +963,24 @@ impl<C: RawComm, T: Clone + Committable + Serialize + DeserializeOwned> Worker<C
...
+            debug!(
+                node     = %self.label,
+                digest   = %digest,
+                present  = %tracker.message.item.is_some(),
+                status   = %tracker.status,
+                ours     = %tracker.ours,
+                elapsed  = ?tracker.start.elapsed(),
+                retries  = %tracker.retries,
+                votes    = %tracker.votes.votes(&digest.commit()),
+                "revisiting ..."
```
jparr721, line 978: Could this perhaps be more clear? At first glance since I haven't touched this yet is that I'm not sure what's going on.

jparr721 (commented):
Appreciate the judicious use of log levels. This is looking good.

cliquenet/src/lib.rs
twittner: From the perspective of cliquenet it is just a warning without further consequences, but for the upper layers it is probably severe. For example if nodes can not send their proposals because they are too large, consensus stalls and can not recover. When the bundle format settles we will have to bring back the size calculation to ensure that messages stay below the max. size (whether it is 5 MiB or something else).

cliquenet/src/lib.rs
twittner: Agreed, it does not have to be a warning.

cliquenet/src/lib.rs
twittner: An invariant of this module is that adding to `task2key` only happens when no prior mapping for the task ID exists. It would be a programmer error if it does, so I think the assertion expresses it well. Would you prefer this slightly longer variant?

```rust
let prev = self.task2key.insert(h.id(), k);
assert!(prev.is_none(), "task to key mapping already exists");
```

sailfish-rbc/src/abraham/worker.rs
twittner: This is near the top of the `retry` method which is invoked with a fixed frequency and revisits all buffered values. What message would you like to see here?

cliquenet/src/lib.rs
twittner: It mirros `IoTask` which has `rh` and `wh` (because it stores 2 `AbortHandle`s).

akonring (approved):
👍

cliquenet/src/lib.rs
```diff
@@ -391,22 +422,33 @@ where
...
+                                continue
+                            };
+                            self.on_connect_task_end(id);
+                            // We only keep the connection if our key is larger than the remote,
+                            // or if we do not have a connection for that key at the moment.
+                            if k < self.key || !self.active.contains_key(&k) {
+                                self.spawn_io(k, s, t)
+                            } else {
+                                debug!(node = %self.key, peer = %k, "dropping new connection");
+                            }
```
akonring, line 445: how do we know that we need to remove only from `task2key` and not also from `connecting`?
Seems simpler to just do:
```suggestion
                            self.on_connect_task_end(id);
                            if let Some(k) = self.lookup_peer(&t) {
                                // We only keep the connection if our key is larger than the remote,
                                // or if we do not have a connection for that key at the moment.
                                if k < self.key || !self.active.contains_key(&k) {
                                    self.spawn_io(k, s, t)
                                } else {
                                    debug!(node = %self.key, peer = %k, "dropping new connection");
                                }
                            }
```

twittner: I simplified that section.

PR #304, opened 2025-04-01
Add the Transaction Generator

jparr721:
Broadcast the same transactions to every node in the committee. Previously each node generated their own transactions.


yapper/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    // Unpack the keyset file which has the urls
+    let keyset = KeysetConfig::read_keyset(&cli.keyset_file).context(format!(
+        "opening the keyfile at path {}",
+        cli.keyset_file.to_string_lossy(),
+    ))?;
+
+    // TODO: make this selection a helper.
+    // Rust is *really* picky about mixing iterators, so we just erase the type.
+    let peer_host_iter: Box<dyn Iterator<Item = &_>> = if cli.multi_region {
```
jparr721, line 55: I'll smooth this out in a subsequent PR. The offset pattern is copied now in 3 spots.

yapper/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+use timeboost::keyset::{KeysetConfig, wait_for_live_peer};
+use timeboost_utils::types::logging::init_logging;
+
+mod tx;
+
+#[derive(Parser, Debug)]
+struct Cli {
+    /// The number of nodes that are being run in this instance.
+    #[clap(long)]
+    nodes: Option<usize>,
```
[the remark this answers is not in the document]
jparr721: Will change this

yapper/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+        // Wait for the peeer to come online so we know it's valid.
+        wait_for_live_peer(address.clone()).await?;
+
+        all_hosts_as_addresses.push(address);
+    }
+
+    // Spawn a new thread per host and let em rip.
+    for address in all_hosts_as_addresses {
+        tokio::spawn(async move {
+            if let Err(err) = tx_sender(cli.tps, address /*KEY */).await {
```
jparr721, line 108: @akonring how can I generate a random key here? I need some decryption key? But that feels wrong. If I make an encrypted bundle as a user, what's the semantics of this operation?

akonring: Yes, in this case the client just uses its encryption public key.
Is it not sufficient to use the same approach as our existing load_generation?
https://github.com/EspressoSystems/timeboost/blob/2c0523ee431fcd779fb60e8d2047236247240770/timeboost-utils/src/load_generation.rs#L24-L30

Setting the `kid` (KeysetId) signals to timeboost that the bundle is encrypted. For now, all key material is static and testing just always uses KeysetId = 1.

twittner (approved):

yapper/src/tx.rs
```diff
@@ -0,0 +1,69 @@
...
+use timeboost_crypto::traits::threshold_enc::ThresholdEncScheme;
+
+use anyhow::{Context, Result};
+use cliquenet::Address;
+use timeboost_utils::load_generation::{make_bundle, tps_to_millis};
+use tokio::time::interval;
+use tracing::{error, warn};
+
+async fn send_transaction(
+    client: Client,
```
twittner, line 14:
```suggestion
    client: &Client,
```

yapper/src/tx.rs
```diff
@@ -0,0 +1,69 @@
...
+    addr: &Address,
+    pubkey: &<DecryptionScheme as ThresholdEncScheme>::PublicKey,
+) -> Result<Response> {
+    let bundle = make_bundle(pubkey)?;
+
+    match bundle {
+        BundleVariant::Regular(bundle) => {
+            let submision_url = Url::parse(&format!("http://{}/v0/submit-regular", addr))
+                .context(format!("parsing {} into a url", addr))?;
```
twittner, line 23: Perhaps the address parsing can happen only once and not for every bundle? Also, to avoid creating the potentially unecessary error string there is `with_context`:
```suggestion
            let submision_url = Url::parse(&format!("http://{}/v0/submit-regular", addr))
                .with_context(|| format!("parsing {} into a url", addr))?;
```

yapper/src/tx.rs
```diff
@@ -0,0 +1,69 @@
...
+use timeboost_crypto::DecryptionScheme;
+use timeboost_crypto::traits::threshold_enc::ThresholdEncScheme;
+
+use anyhow::{Context, Result};
+use cliquenet::Address;
+use timeboost_utils::load_generation::{make_bundle, tps_to_millis};
+use tokio::time::interval;
+use tracing::{error, warn};
+
+async fn send_transaction(
```
twittner, line 13: I think `send_transaction` can be inlined in `tx_handler` and then simplified a little.

yapper/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+            let pubkey = keyset.dec_keyset().pubkey()?;
+            async move {
+                if let Err(err) = tx_sender(cli.tps, address, pubkey).await {
+                    error!(%err, "tx sender failed");
+                }
+            }
+        }));
+    }
+
+    let mut signal = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
```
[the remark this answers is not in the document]
lukeiannucci: yea that makes sense

scripts/runit
```diff
@@ -111,7 +111,33 @@ for (( i=0; i<$nodes; i++ )); do
...
+        echo "Process $pid completed successfully with exit code $exit_code."
+    else
+        echo "Process $pid failed with exit code $exit_code."
+    fi
+done
+
+if [ -n "$tps" ] && [ "$tps" -gt 0 ]; then
+    echo "Attempting to kill yapper..."
+    pkill yapper
+fi
```
twittner, line 142: Would this not suffice?
```suggestion
if [ "$tps" -gt 0 ]; then
    target/release/yapper --tps $tps --nodes $nodes --keyset-file $keyset_file &
fi

wait "${pids[@]}"

if [ "$tps" -gt 0 ]; then
    pkill yapper
fi
```

lukeiannucci: yep that is a lot cleaner

docker/yapper.Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Builder stage
+FROM rust:bullseye AS builder
+
+WORKDIR /app
+
+COPY . .
+
+RUN cargo install just
+
+RUN just build_release
```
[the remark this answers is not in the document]
lukeiannucci: yea probably not, i just copy and pasted timeboost.Dockerfile, should have paid closer attention

docker/yapper.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Copy binary and just
+COPY --from=builder /app/target/release/yapper .
+COPY --from=builder /app/test-configs .
+
+# Set ownership of application files and make binary executable
+RUN chown -R yapperuser:appgroup /app && chmod +x /app/yapper
+
+# We need curl for the healthcheck
+RUN apt update && apt install -yqq curl
```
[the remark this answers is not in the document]
lukeiannucci: i will verify

yapper/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+            ))?
+            .parse::<u16>()
+            .context("parsing port into u16")?;
+        let mut address = Address::from((host, port));
+
+        // Wait for the peeer to come online so we know it's valid.
+        wait_for_live_peer(address.clone()).await?;
+        address.set_port(800 + address.port());
+        all_hosts_as_addresses.push(address);
+    }
```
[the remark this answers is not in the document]
lukeiannucci: yes this is super hacky, if you look in `wait_for_live_peer()` it is also doing the hack of adding +800 to the port to get rpc port. i will make a follow up PR to this and make it cleaner

docker/yapper.Dockerfile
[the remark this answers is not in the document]
lukeiannucci: i think this is needed for probing the peer addresses when we call `wait_for_live_peer()` and the creating of our `request::client()`. otherwise the node will container will crash

yapper/src/tx.rs
lukeiannucci: https://github.com/EspressoSystems/timeboost/pull/304/commits/1b873f4e5e1c698176fb57b6664534387abe1619

yapper/src/tx.rs
lukeiannucci: lots of refactoring done:
https://github.com/EspressoSystems/timeboost/pull/304/commits/1b873f4e5e1c698176fb57b6664534387abe1619

yapper/src/tx.rs
```diff
@@ -0,0 +1,69 @@
...
+        .build()
+        .context("building the reqwest client")?;
+    loop {
+        tokio::select! {
+            _ = interval.tick() => {
+                if let Err(err) = send_transaction(client.clone(), &addr, &pubkey).await {
+                    error!(%err, "failed to send transaction");
+                }
+            }
+            _ = tokio::signal::ctrl_c() => {
```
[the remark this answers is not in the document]
lukeiannucci: https://github.com/EspressoSystems/timeboost/pull/304/commits/1b873f4e5e1c698176fb57b6664534387abe1619

yapper/src/main.rs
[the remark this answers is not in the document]
lukeiannucci: https://github.com/EspressoSystems/timeboost/pull/304/commits/1b873f4e5e1c698176fb57b6664534387abe1619

timeboost-utils/Cargo.toml
```diff
@@ -17,6 +17,8 @@ bincode = { workspace = true }
 blake3 = { workspace = true }
 bs58 = { workspace = true }
 bytes = { workspace = true }
+cliquenet = { path = "../cliquenet" }
+dotenvy = { workspace = true }
```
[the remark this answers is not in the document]
lukeiannucci: https://github.com/EspressoSystems/timeboost/pull/304/commits/1b873f4e5e1c698176fb57b6664534387abe1619

yapper/src/tx.rs
lukeiannucci: @twittner - Actually i noticed that the way we were generating the bundles was incorrect. As each `tx_sender` task was generating its own bundle and sending it to its own node. So if my understanding is correct the whole purpose of this PR was done incorrectly. I did a lot of refactoring to send the same bundle to each node, you might need to take another close look at this commit:
https://github.com/EspressoSystems/timeboost/pull/304/commits/1b873f4e5e1c698176fb57b6664534387abe1619

yapper/src/main.rs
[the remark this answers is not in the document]
lukeiannucci: created task here:
https://app.asana.com/1/1208976916964769/project/1209401855772422/task/1210200948542497

docker/yapper.Dockerfile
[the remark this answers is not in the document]
twittner: Is `wait_for_live_peer` not just using `reqwest`? Where is the dependency to `curl`?

yapper/src/tx.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    Ok(clients_and_urls)
+}
+
+async fn send_bundle_to_node(
+    bundle: &BundleVariant,
+    client: &Client,
+    regular_url: &str,
+    priority_url: &str,
```
twittner, line 37: Since the URLs have been parsed in `setup_clients_and_urls`, why not accepting them here as `Url` parameters?

yapper/src/tx.rs
```diff
@@ -0,0 +1,87 @@
...
+        let priority_url = Url::parse(&format!("http://{}/v0/submit-priority", addr))
+            .with_context(|| format!("parsing {} into a url", addr))?;
+
+        clients_and_urls.push((client, regular_url, priority_url));
+    }
+
+    Ok(clients_and_urls)
+}
+
+async fn send_bundle_to_node(
```
twittner, line 33: I think this one should also inspect the response status and error if not 200, e.g.

```rust
    let response = match bundle {
        BundleVariant::Regular(bundle) => client
            .post(regular_url)
            .json(&bundle)
            .send()
            .await?,
        BundleVariant::Priority(signed_priority_bundle) => client
            .post(priority_url)
            .json(&signed_priority_bundle)
            .send()
            .await?
    };
    if !response.status().is_success() {
        return Err(anyhow!("response status: {}", response.status()))
    }
    Ok(())
```

yapper/src/tx.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use anyhow::{Context, Result};
+use cliquenet::Address;
+use tracing::{error, warn};
+
+fn setup_clients_and_urls(all_hosts_as_addresses: &[Address]) -> Result<Vec<(Client, Url, Url)>> {
+    let mut clients_and_urls = Vec::new();
+
+    for addr in all_hosts_as_addresses {
+        // Create a new client for each address
```
twittner, line 16: There is no need to create a reqwest `Client` for each address. A single client can be shared (it has its own internal connection pool).

yapper/src/tx.rs
lukeiannucci: i was just trying to avoid `url.clone()`

twittner: The clone should be cheaper than parsing again I reckon.

docker/yapper.Dockerfile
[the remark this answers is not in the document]
lukeiannucci: yea you were right, the issue was it didn't need dependency to curl, after i went through the logs. it looks like it needs a openssl.

yapper/src/tx.rs
lukeiannucci: https://github.com/EspressoSystems/timeboost/pull/304/commits/c484a491628d4471c7e268b10f0d3fcd9123579d

yapper/src/tx.rs
```diff
@@ -0,0 +1,87 @@
...
+            warn!("failed to generate bundle");
+            continue;
+        };
+
+        interval.tick().await;
+
+        join_all(urls.iter().map(|(regular_url, priority_url)| async {
+            send_bundle_to_node(&b, &c, regular_url, priority_url).await
+        }))
+        .await;
```
twittner, line 83: This will return a vector of `Result`s. Should this not be handled? If not, perhaps `send_bundle_to_node` should return unit instead?

yapper/src/tx.rs
lukeiannucci: https://github.com/EspressoSystems/timeboost/commit/c484a491628d4471c7e268b10f0d3fcd9123579d

yapper/src/tx.rs
lukeiannucci: i handle the logging in the `send_bundle_to_node`. i can handle it here, i originally had what you commented, but the only thing that would come out of this is logging. as i assume if we fail to send to a node, we would just keep the loop going instead of breaking out.

twittner: > perhaps `send_bundle_to_node` can just have no return value?

:+1:

lukeiannucci: https://github.com/EspressoSystems/timeboost/pull/304/commits/4b2e714d40f1b32b87638c8c54f4914ea2c7cc3e

yapper/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let nodes = select_peer_hosts(keyset.keyset(), cli.nodes, cli.multi_region);
+
+    let mut addresses = Vec::new();
+    for node in nodes {
+        info!("waiting for peer: {}", node.sailfish_url);
+        let mut addr = node
+            .sailfish_url
+            .parse::<Address>()
+            .expect("failed to parse saiflish url to address");
```
twittner, line 67:
```suggestion
            .context("failed to parse saiflish url to address")?;
```

PR #307, opened 2025-04-02
Check that vote sender is vote signer.

twittner:
Just like with proposals it is an error to send votes on behalf of other parties.


akonring (approved):
LGTM.

PR #308, opened 2025-04-03
Ensure deduplication cache invariants.

twittner:
During the inclusion phase, the cache must end with at least 8 consecutive rounds to allow a node to participate in subsequent phases.

Edit: This also changes the way, Sailfish is allowed to get ahead of Timeboost.


akonring (approved):
LGTM.

PR #309, opened 2025-04-04
Reconnect

twittner:
Currently, when the buffer to a peer overflows, a new connect attempt is started. The intention is that this new connection (once established) replaces the current one. However, if the public key of the reconnecting node is smaller than the remote one, the peer will drop the accepted connection in preference of the currently active one. To enforce a reconnect, this commit first drops the active connection before connecting again.

The second commit adds a message size histogram metric.


jparr721 (approved):
🎉

PR #310, opened 2025-04-07
Fix priority bundles

akonring:
**First PR in a series of upcoming changes.** This temporarily disables encrypted load_gen to enable downstream fixes before re-enabling encryption.

## Context

If a priority bundle isn’t included—e.g., because another bundle with the same sequence number but lower hash takes precedence—it gets added to the retry list. These retried bundles are then blindly appended to the bundle queue in the next round.

The spec touches on this behavior [here](https://github.com/OffchainLabs/decentralized-timeboost-spec/blob/e314d70f9d0ebabccea17a59804e3b90ec96b936/inclusion.md?plain=1#L152), noting that the retry mechanism helps ensure eventual inclusion.

The core issue is that retried bundles are naively re-added, regardless of conflicts at the same sequence number. This causes Sailfish to accumulate increasingly large vertices until it either fails or a new epoch resets the priority queue.

### Load Generator as Amplification

Profiling Timeboost on a test like transaction-order—which uses large bundles—quickly reveals the issue: most of the time is spent creating and committing vertices.

Additionally, because priority bundles are distributed in the same way as non-priority bundles (i.e., broadcast to all nodes), the problem is amplified. This results in an even greater buildup of bundles with duplicate sequence numbers.

### Relation to the Coupon Collector’s Problem

So did this ever work? Have Timeboost ever produced inclusion lists with priority bundles?

Yes—despite the underlying issue, Timeboost has occasionally produced inclusion lists containing priority bundles. In particular, the transaction-order test would sometimes manage to push priority bundles through, triggering transient errors in downstream logic such as decryption.

Why only occasionally? In the test, load_gen submits priority bundles with probability $p=0.1$ over 500 transactions, yielding around 50 priority candidates. However, due to how Transaction::arbitrary assigns timestamps—targeting the current epoch only 50% of the time—only about half (≈25) are actually eligible for inclusion in the current round.

Now, what’s the chance these ~25 transactions cover all sequence numbers in the target set {0,...,10}? This is a variation of the [Coupon Collector’s Problem](https://en.wikipedia.org/wiki/Coupon_collector%27s_problem), where the expected number of samples needed to collect all $n$ unique items is $\Theta(n \log n)$. 
For $n=11$, the expected number is approximately 26.

So with just 25 eligible priority bundles per test run, it’s unsurprising that full coverage—and thus successful inclusion—only happens occasionally.

### Why do we only see this issue now

Previously, the load generator submitted transactions and priority bundles with payloads around 8 bytes, allowing Sailfish to run longer before hitting issues with large vertices. Now that bundle sizes are in the 100–200 byte range, the problem surfaces much more quickly—often within just a few rounds.

While tests like transaction-order verify that Timeboost nodes agree on the produced transactions and bundles, they don’t check whether the output matches the load that was generated. A test validating this alignment would have caught the bug earlier.

Lastly, the shift to a bundle-centric model introduced a related bug in load_gen: it only generated priority bundles with seqno in  {0,...,10}. As a result, these bundles consistently failed verification, since the inclusion phase expects sequence numbers to start at 0 in each epoch.

## Rationale

This PR introduces sequence number deduplication on the queuing phase, before consensus. While no rational client should submit multiple bundles with the same sequence number and epoch, if it does happen, this change ensures deduplication happens early—when forming the candidate list—rather than after consensus. This prevents Sailfish from wasting resources processing duplicate data that would ultimately be filtered out during final inclusion list construction anyway.

## Content

The PR contains the following changes:

1. Removes encrypted bundle logic from load generation (see justification above).

2. Applies deduplication to retried bundles by storing priority bundles in a BTreeMap keyed by sequence number.

3. Fixes a bug in load generation to ensure priority bundle sequence numbers are sampled uniformly from the range [0, 10], inclusive.

### DIFF
--- a/timeboost-sequencer/src/queue.rs
+++ b/timeboost-sequencer/src/queue.rs
@@ -143,21 +143,22 @@ impl BundleQueue {
         }
 
         for b in priority {
-            inner
-                .priority
-                .entry(b.bundle().epoch())
-                .and_modify(|bundles| match bundles.entry(b.seqno()) {
+            match inner.priority.entry(b.bundle().epoch()) {
+                Entry::Occupied(mut bundles) => match bundles.get_mut().entry(b.seqno()) {
                     Entry::Vacant(e) => {
-                        e.insert(b.clone());
+                        e.insert(b);
                     }
                     Entry::Occupied(mut e) => {
                         let sb = e.get_mut();
                         if b.digest() < sb.digest() {
-                            *sb = b.clone();
+                            *sb = b;
                         }
                     }
-                })
-                .or_insert([(b.seqno(), b)].into());
+                },
+                Entry::Vacant(bundles) => {
+                    bundles.insert([(b.seqno(), b)].into());
+                }
+            }
         }
 
         inner
--- a/timeboost-sequencer/src/queue.rs
+++ b/timeboost-sequencer/src/queue.rs
@@ -1,10 +1,13 @@
+use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, VecDeque};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use parking_lot::Mutex;
 use sailfish::types::{DataSource, RoundNumber};
-use timeboost_types::{Address, Bundle, BundleVariant, Epoch, RetryList, SignedPriorityBundle};
+use timeboost_types::{
+    Address, Bundle, BundleVariant, Epoch, RetryList, SeqNo, SignedPriorityBundle,
+};
 use timeboost_types::{CandidateList, DelayedInboxIndex, InclusionList, Timestamp};
 use tracing::trace;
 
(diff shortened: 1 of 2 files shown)


timeboost-sequencer/src/queue.rs
```diff
@@ -128,14 +128,24 @@ impl BundleQueue {
             inner
                 .priority
                 .entry(b.bundle().epoch())
-                .or_default()
-                .push(b)
+                .and_modify(|bundles| match bundles.entry(b.seqno()) {
+                    Entry::Vacant(e) => {
+                        e.insert(b.clone());
```
twittner, line 133: To avoid the cloning of bundles, perhaps something like this could work:

```rust
for b in priority {
    match inner.priority.entry(b.bundle().epoch()) {
        Entry::Occupied(mut bundles) => {
            match bundles.get_mut().entry(b.seqno()) {
                Entry::Vacant(e) => {
                    e.insert(b);
                }
                Entry::Occupied(mut e) => {
                    let sb = e.get_mut();
                    if b.digest() < sb.digest() {
                        *sb = b;
                    }
                }
            }
        }
        Entry::Vacant(bundles) => {
            bundles.insert([(b.seqno(), b)].into());
        }
    }
}
```

akonring: Yes, that is a great idea. Thanks.

PR #313, opened 2025-04-07
Fix decryption

akonring:
### Content
This PR fixes the decrypter logic and re-enables load generation with encrypted bundles.

1. The cid2idx mapping was stored using a BiMap. This caused entries to be overwritten and would eventually corrupt the state of the incubator.
2. The decrypter would insert shares from other nodes even if it had already hatched the corresponding ciphertext, effectively, causing the node to "start over" and eventually failing to decrypt because of missing shares.
3. The assembly of inclusion list would index into the decrypted data using the total number of priority bundles (instead of the number of modified priority bundles).

This PR fixes the above.


akonring:
superseded by #315

PR #314, opened 2025-04-07
Use `CandidateListBytes` to estimate message size.

twittner:
The network layer has a message size limit which requires the sequencer to produce data items that do not cause the resulting vertex message to exceed the allowed maximum. Since the message size is dependent on the encoding format, this commit changes the payload data type from `CandidateList` to `CandidateListBytes` (a newtype around `Bytes`) and always serializes to bincode. When building the candidate list it uses bincode's `SizeWriter` to calculate an estimate.


timeboost-sequencer/src/queue.rs
```diff
@@ -181,26 +189,66 @@ impl DataSource for BundleQueue {
...
+
+fn bincode_len<T>(val: T) -> Result<usize, bincode::error::EncodeError>
+where
+    T: serde::Serialize,
+{
+    use bincode::config::standard;
+    use bincode::enc::write::SizeWriter;
+    use bincode::serde::BorrowCompat;
+
+    let mut w = SizeWriter::default();
```
akonring, line 251: Thought: presumably we could avoid allocating new writers here by passing in the full list. But the current approach still seems cleaner, imo.

akonring (approved):
LGTM.

timeboost-sequencer/src/queue.rs
twittner: No need to worry, [`SizeWriter`](https://docs.rs/bincode/latest/bincode/enc/write/struct.SizeWriter.html) is just a newtype around `usize`.

PR #315, opened 2025-04-08
Decryption phase with catchup support

akonring:
EDIT: The use of RBC has been replaced by a custom reliable layer in cliquenet which renders most of the considerations below moot.

### Content
This PR replaces the existing decryption network with an RBC-based implementation (using best-effort broadcast). It addresses bugs in the decrypter logic that surfaced after PRs #298, #310, and #312, and re-enables load generation with encrypted bundles.

### Main Design Change
The decryption phase now uses the Rbc abstraction, leveraging best-effort broadcast. As a consequence, some decryption-specific types are now exposed to Sailfish crates.

### Future Design Considerations
For robustness—especially in scenarios involving late starts or catch-up—it’s important to support ACKs. The current solution uses only the minimal portion of Rbc needed. Ideally, ACKs and basic serialization should be abstracted into a reusable layer beneath the Abraham et al. protocol, which could serve as a dedicated decryption network.

After discussions with OCL, we confirmed that the same machines serve as both the consensus and decryption committees. This reopens the design space to revisit the original idea of using a single unified network for Sailfish and the broader Timeboost system.

### Miscellaneous Fixes
The cid2idx mapping used a BiMap, which led to overwritten entries and state corruption in the incubator.

The decrypter accepted shares even after the corresponding ciphertext had hatched, effectively restarting progress and eventually failing due to missing shares.

The inclusion list was incorrectly indexed using the total number of priority bundles, rather than the number of modified ones.

### DIFF
--- a/timeboost-sequencer/src/decrypt.rs
+++ b/timeboost-sequencer/src/decrypt.rs
@@ -345,16 +345,13 @@ impl Worker {
                             }
                             if hatched_rounds.is_empty() {
                                 // fast-forward
-                                if let Some(last_round) = self
-                                    .shares
-                                    .keys()
-                                    .filter(|k| k.round() < round)
-                                    .max()
-                                {
-                                    // mark all earlier rounds as hatched
-                                    let removed = self.shares.split_off(&last_round.clone());
-                                    hatched_rounds.extend(removed.keys().map(|k| k.round()));
-                                }
+                                self.shares.retain(|k, _| {
+                                    let old = k.round() < round;
+                                    if old {
+                                        hatched_rounds.insert(k.round());
+                                    }
+                                    !old
+                                });
                             }
                         }
                         Err(e) => {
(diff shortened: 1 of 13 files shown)


timeboost-sequencer/src/decrypt.rs
```diff
@@ -442,23 +493,37 @@ impl Worker {
...
 
         Ok(Some((round, decrypted.into_values().collect())))
     }
 }
 
+/// Serialize a given data type into `Bytes`
+fn serialize<T: Serialize>(d: &T) -> Result<bytes::Bytes> {
+    let mut b = BytesMut::new().writer();
+    bincode::serde::encode_into_std_write(d, &mut b, bincode::config::standard())?;
+    Ok(b.into_inner().freeze())
```
twittner, line 517:
```suggestion
    Ok(b.into_inner())
```
If `serialize` returns `BytesMut` it could be passed directly to the network.

twittner (approved):
Looks good to me.

timeboost-sequencer/src/decrypt.rs
akonring: nice, thanks.

timeboost-sequencer/src/decrypt.rs
```diff
@@ -366,21 +419,21 @@ impl Worker {
...
-    async fn broadcast(&self, share: &ShareInfo) -> Result<(), DecryptError> {
-        let share_bytes = bincode::serde::encode_to_vec(share, bincode::config::standard())
-            .map_err(DecryptError::BincodeEncode)?;
+    async fn broadcast(&mut self, share_info: &ShareInfo) -> Result<()> {
+        let share_bytes = serialize(share_info)?;
         self.net
-            .multicast(share_bytes.into())
+            .send(None, share_bytes)
             .await
+            .map(|_| ())
```
twittner, line 427: Eventually we need to store the `SeqId` and perform garbage collection in the network when rounds are no longer relevant.

akonring: Yes, agree that this is important in practice but will leave it out of scope for this PR.
In general it seems that this garbage collection should align with Sailfish gc window as otherwise nodes that are catching up will not be able to decrypt and move forward.

PR #321, opened 2025-04-15
Garbage collect decryption network messages

akonring:
This PR adds garbage collection (GC) for messages in the overlay network used during the decryption phase.

For catch-up nodes, it's crucial that the decryption phase's GC window is at least as large as Sailfish's GC window. Otherwise, Sailfish might generate inclusion lists that cannot be decrypted because the corresponding messages with decryption shares have already been garbage collected. To prevent this, the decryption phase now depends on Sailfish's GC round. However, just because a round can be GC’ed in Sailfish doesn’t mean it should be GC’ed in the decryption phase—especially when the decryption phase is lagging behind. In such cases, prematurely GC’ing may drop messages still needed by other nodes for decryption.

To address this, the decryption GC window is defined relative to Sailfish GC round (`sailfish_gc_round - 100`). The constant 100 is already somewhat heuristically applied as the channel capacity between the Decrypter and its Worker, effectively acting as a leash between Sailfish and Timeboost. Since the Timeboost sequencer checks this channel’s capacity before allowing Sailfish to proceed, this value provides a safe margin for GC coordination.

### DIFF
--- a/timeboost-sequencer/src/decrypt.rs
+++ b/timeboost-sequencer/src/decrypt.rs
@@ -88,14 +88,17 @@ impl Decrypter {
         self.dec_tx.capacity() > 0 && self.enc_tx.capacity() > 0
     }
 
+    pub async fn gc(&mut self, r: RoundNumber) -> Result<()> {
+        self.enc_tx
+            .send(WorkerCommand::Gc(r))
+            .await
+            .map_err(|_| DecryptError::Shutdown)
+    }
+
     /// Identifies encrypted bundles in inclusion lists,
     /// computes the expected state diff, then sends the
     /// encrypted data to the worker for hatching.
-    pub async fn enqueue(
-        &mut self,
-        gc_round: Option<RoundNumber>,
-        incl: InclusionList,
-    ) -> Result<()> {
+    pub async fn enqueue(&mut self, incl: InclusionList) -> Result<()> {
         let round = incl.round();
         let total_items = incl.len();
         let (encrypted_pb_idx, encrypted_pb_data): (Vec<_>, Vec<_>) = incl
@@ -143,12 +146,6 @@ impl Decrypter {
             self.incls.insert(round, Status::Encrypted(incl));
             self.modified
                 .insert(round, (encrypted_pb_idx, encrypted_rb_idx));
-        }
-        if let Some(r) = gc_round {
-            self.enc_tx
-                .send(WorkerCommand::Gc(r))
-                .await
-                .map_err(|_| DecryptError::Shutdown)?;
         }
         trace!(
             node   = %self.label,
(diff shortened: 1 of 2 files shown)


timeboost-sequencer/src/decrypt.rs
```diff
@@ -264,6 +281,7 @@ struct Worker {
     dec_sk: DecryptionKey,
     cid2idx: HashMap<Nonce, usize>,
     cid2ct: BiMap<(RoundNumber, Nonce), Ciphertext>,
+    round2seqs: BTreeMap<RoundNumber, Vec<SeqId>>,
```
twittner, line 284: Is there a need to store all sequence IDs of a round? Why not just the last one?

timeboost-sequencer/src/decrypt.rs
```diff
@@ -328,55 +343,68 @@ impl Worker {
...
+                                Err(e) => {
+                                    warn!("failed to decrypt data: {:?}", e);
+                                    continue;
+                                }
                             }
-                        }
-                        Err(e) => {
-                            warn!("failed to decrypt data: {:?}", e);
+                        },
+                        WorkerCommand::Gc(round) => {
```
twittner, line 384: I think this also needs to do invoke `Overlay::gc` with the last sequence Id corresponding to that round number.

timeboost-sequencer/src/decrypt.rs
```diff
@@ -328,55 +343,68 @@ impl Worker {
...
-                                warn!("failed to insert local shares: {:?}", e);
-                                continue;
-                            }
-                            if catching_up {
-                                // fast-forward
-                                self.shares.retain(|k, _| {
-                                    let old = k.round() < round;
-                                    if old {
-                                        hatched_rounds.insert(k.round());
+                Some(command) = enc_rx.recv() => {
```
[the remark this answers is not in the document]
akonring: thanks for pointing this out. fixed.

timeboost-sequencer/src/decrypt.rs
```diff
@@ -427,11 +455,15 @@ impl Worker {
...
-            .map_err(DecryptError::net)
+        let b = self.net.broadcast(share_bytes).await;
+        if let Ok(seqid) = b {
+            self.round2seqs
+                .entry(share_info.round())
+                .or_default()
+                .push(seqid);
+            return Ok(());
+        }
+        b.map(|_| ()).map_err(DecryptError::net)
```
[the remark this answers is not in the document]
akonring: fixed.

timeboost-sequencer/src/decrypt.rs
akonring: No. Changed it to only store the last seq id for the decryption round.

timeboost-sequencer/src/decrypt.rs
akonring: Yeah, that part got lost somehow. Should be fixed now.

PR #322, opened 2025-04-17
Collect round info when recovering. 

twittner:
Implements Fig. 11 of the spec. One complication is how to know when to recover. This PR solves this by creating a stamp file when first starting. Subsequent starts will switch to recovery when the file exists. The command-line option `--stamp` has been added to allow specifying the path of the stamp file. To make test scripts not terribly inconvenient to use a second option `--ignore-stamp` is available which does what it says.

In addition to that `Evidence::is_valid` has been improved by always requiring a round number as parameter, as the validity of the evidence is linked to the round number it is meant to support.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -90,8 +90,8 @@ jobs:
         tar -vxf just-1.14.0-x86_64-unknown-linux-musl.tar.gz just
         sudo cp just /usr/bin/just
     - name: Run tests
-      run: just run_demo -url http://localhost:8547
+      run: just run_demo -url http://localhost:8547 -s /tmp/stamp --ignore-stamp
     - name: Run tests with late-start node
-      run: just run_demo -l -url http://localhost:8547
+      run: just run_demo -l -url http://localhost:8547 -s /tmp/stamp --ignore-stamp
     - name: Run sailfish demo
       run: just run_sailfish_demo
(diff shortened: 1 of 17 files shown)


sailfish-types/src/message.rs
```diff
@@ -374,11 +361,11 @@ impl Evidence {
         }
     }
 
-    pub fn is_valid(&self, c: &Committee) -> bool {
+    pub fn is_valid(&self, r: RoundNumber, c: &Committee) -> bool {
```
akonring, line 364: Thought: Seems a bit weird that `is_valid` should take a round as argument when it evidence is always apply to the immediate predecessor.

akonring (approved):
Nice.

sailfish-types/src/message.rs
twittner: This method establishes the fact that the evidence applies to the correct round number which is not a given. Previously this check had to happen in addition to calling `Evidence::is_valid` which is repetitive and error prone as it can easily be forgotten.

.github/workflows/build-and-test.yml
```diff
@@ -90,8 +90,8 @@ jobs:
         tar -vxf just-1.14.0-x86_64-unknown-linux-musl.tar.gz just
         sudo cp just /usr/bin/just
     - name: Run tests
-      run: just run_demo -url http://localhost:8547
+      run: just run_demo -url http://localhost:8547 -s /tmp/stamp --ignore-stamp
```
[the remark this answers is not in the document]
twittner: The default requires `-s | --stamp`. We could omit `--ignore-stamp` but given that we call `run_demo` multiple times I wanted to make sure we do not attempt recovery (which would fail when all nodes try to recover).

PR #323, opened 2025-04-18
Use `Bucket`s in `Overlay`.

twittner:
Due to message reordering the use of `SeqId` for garbage collection is not correct. If for example a proposal in round `r + 1` arrived and we send back a vote, yielding an `x : SeqId` and we later send out a proposal for a round `r`, we get a `y : SeqId` with `y > x`. When garbage collecting `r` we would remove all sequence numbers up to `y` which includes `x` even though round `r + 1` is still active.

Here we change the approach and require senders to give us a bucket number (usually corresponding directly to a round number) when sending a message and also when garbage collecting.

### DIFF
--- a/cliquenet/src/overlay.rs
+++ b/cliquenet/src/overlay.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::convert::Infallible;
 use std::ops::Deref;
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


akonring (approved):
This is a really solid approach. The only consideration is that the semantics might get a bit tricky once we start setting up buckets for the Block production phase. In that context, we don't have a concept of a round number — instead, we use BlockNumber, which might overlap with round numbers but has different meaning. If needed, we could encode this distinction directly into the bucket.

PR #324, opened 2025-04-18
Use `bincode::Configuration::with_limit`.

twittner:
Otherwise malformed input can cause it to allocate too much, e.g. something like this:

```rust
let input = [253, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF];
let _ = bincode::decode_from_slice::<Vec<u8>, _>(&input, bincode::config::standard());
```

will result in a panic ("capacity overflow").

### DIFF
--- a/timeboost-crypto/src/lib.rs
+++ b/timeboost-crypto/src/lib.rs
@@ -3,10 +3,11 @@ pub mod sg_encryption;
 pub mod traits;
 
 use ark_ec::CurveGroup;
-use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize, SerializationError};
 use committable::{Commitment, Committable, RawCommitmentBuilder};
 use cp_proof::Proof;
 use digest::{generic_array::GenericArray, typenum};
+use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use sg_encryption::ShoupGennaro;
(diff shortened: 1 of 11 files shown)


timeboost-crypto/src/lib.rs
```diff
@@ -156,90 +157,68 @@ impl From<Nonce> for GenericArray<u8, typenum::U12> {
...
-            .map(|(val, _)| val)
-            .map_err(|_| ark_serialize::SerializationError::InvalidData)
+    pub fn try_from_str<const N: usize>(value: &str) -> Result<Self, SerializationError> {
+        try_from_str::<Self, N>(value)
     }
 }
 
 impl<C: CurveGroup> KeyShare<C> {
-    pub fn as_bytes(&self) -> Vec<u8> {
+    pub fn to_bytes(&self) -> Vec<u8> {
```
akonring, line 190: thanks 👍

timeboost/src/keyset.rs
```diff
@@ -84,10 +84,10 @@ pub fn private_keys(
         let bytes = &bs58::decode(dec_key)
             .into_vec()
             .context("unable to decode bs58")?;
-        let dec_key: KeyShare =
-            bincode::serde::decode_from_slice(bytes, bincode::config::standard())
-                .map(|(val, _)| val)
-                .expect("unable to read bytes into keyshare");
+        let config = bincode::config::standard().with_limit::<8192>();
```
akonring, line 87: Thought: Some places the level of granularity seems a bit unnecessary. Would it make sense to use a shared bincode config for most encode/decode cases — similar to how [Hotshot does it](https://github.com/EspressoSystems/espresso-network/blob/e70041f969c4cd75073d5e45ceb493963b216604/crates/hotshot/types/src/utils.rs#L316)?

akonring (approved):
LGTM.
More robust handling of malicious input is probably something we'll need to tackle more broadly in the future — and at that point, we may want to reconsider using bincode. For now, limiting the maximum memory allocation during deserialization is a solid safeguard against things like maliciously large length headers.

PR #325, opened 2025-04-18
Block production init

akonring:
### Context
This PR introduces the Block Production Phase, which consumes a decrypted, timestamped transaction stream and produces deterministic, certified blocks. Once a block is constructed, the node signs and broadcasts it. Upon receiving $$t+1$$ signatures (including at least one honest signature), the block is considered certified.

### What’s Included
**Multiplexing Network Messages**
To avoid spinning up a separate Timeboost network instance for the Block Production Phase, this PR introduces a Multiplex component. It routes incoming Timeboost messages to the appropriate consumer (Decryption or Block Production), ensuring clean separation between protocol phases over a shared network layer.

**Block Production**
This PR adds a simple but deterministic block construction algorithm that batches exactly 10 transactions per block. A minimal Arbitrum/Ethereum-like block structure is created, including fields like parent, txs, and tx_root. The implementation references existing centralized sequencer code [here](https://github.com/EspressoSystems/timeboost/blob/65efbb07f6c27fd7dbada049a6d7ac7af33c4fea/timeboost-sequencer/src/produce.rs#L64-L67), offering context for future spec-aligned implementations.

### Additional Notes
~~**Garbage Collection**
GC is currently not implemented due to the incomplete GC interface in the Overlay network. As a result, un-ACKed Overlay messages are not garbage collected. This has no impact on current tests or demos. TODOs are left in the relevant places for future GC integration once the interface stabilizes.~~

**Timestamps & Block Format**
Per the [decentralized-timeboost spec](https://github.com/OffchainLabs/decentralized-timeboost-spec/blob/e314d70f9d0ebabccea17a59804e3b90ec96b936/README.md?plain=1#L166), each transaction now includes a timestamp in the Sorting Phase. While the current block construction doesn't yet use these timestamps, the groundwork is laid for future alignment with the finalized block format.

**Transaction Order Test**
The transaction-order test has been updated to reflect the new block-based output. It now verifies that all nodes produce identical certified blocks with matching hashes.

### DIFF
--- a/multisig/src/envelope.rs
+++ b/multisig/src/envelope.rs
@@ -71,6 +71,15 @@ impl<D: Committable, S> Envelope<D, S> {
     }
 }
 
+impl<D: Committable> From<Envelope<D, Validated>> for Envelope<D, Unchecked> {
+    fn from(value: Envelope<D, Validated>) -> Self {
+        Envelope {
+            signed: value.signed,
+            _marker: PhantomData,
+        }
+    }
+}
+
 impl<D: Committable, S> Deref for Envelope<D, S> {
     type Target = Signed<D>;
 
--- a/multisig/src/envelope.rs
+++ b/multisig/src/envelope.rs
@@ -71,15 +71,6 @@ impl<D: Committable, S> Envelope<D, S> {
     }
 }
 
-impl<D: Committable> From<Envelope<D, Validated>> for Envelope<D, Unchecked> {
-    fn from(value: Envelope<D, Validated>) -> Self {
-        Envelope {
-            signed: value.signed,
-            _marker: PhantomData,
-        }
-    }
-}
-
 impl<D: Committable, S> Deref for Envelope<D, S> {
     type Target = Signed<D>;
 
--- a/timeboost-sequencer/src/lib.rs
+++ b/timeboost-sequencer/src/lib.rs
@@ -35,6 +35,8 @@ use sort::Sorter;
 type Result<T> = std::result::Result<T, TimeboostError>;
 type Candidates = VecDeque<(RoundNumber, Vec<CandidateList>)>;
 
+const MAX_SIZE: usize = 100;
+
 #[derive(Debug)]
 pub struct SequencerConfig {
     priority_addr: Address,
--- a/timeboost-sequencer/src/lib.rs
+++ b/timeboost-sequencer/src/lib.rs
@@ -2,7 +2,6 @@ mod decrypt;
 mod include;
 mod metrics;
 mod multiplex;
-#[allow(dead_code)]
 mod produce;
 mod queue;
 mod sort;
@@ -193,11 +192,12 @@ impl Sequencer {
         queue.set_max_data_len(cliquenet::MAX_MESSAGE_SIZE - 128 * 1024);
 
         // Demultiplexing of Timeboost network messages.
-        let (dec_rx, block_rx, multiplex) =
-            Multiplex::go(public_key, committee.clone(), Overlay::new(network));
+        let multiplex = Multiplex::new(public_key, committee.clone(), Overlay::new(network));
+        let multiplex_tx = multiplex.tx();
+        let (dec_rx, block_rx) = multiplex.go();
 
         let decrypter =
-            Decrypter::new(public_key, keyset, cfg.dec_sk, dec_rx, multiplex.tx.clone());
+            Decrypter::new(public_key, keyset, cfg.dec_sk, dec_rx, multiplex_tx.clone());
 
         let (tx, rx) = mpsc::channel(1024);
 
@@ -208,7 +208,7 @@ impl Sequencer {
             includer: Includer::new(committee.clone(), cfg.index),
             decrypter,
             sorter: Sorter::new(),
-            producer: BlockProducer::new(cfg.keypair, committee, block_rx, multiplex.tx.clone()),
+            producer: BlockProducer::new(cfg.keypair, committee, block_rx, multiplex_tx),
             output: tx,
             mode: Mode::Passive,
         };
@@ -306,7 +306,10 @@ impl Task {
                 result = self.decrypter.next() => match result {
                     Ok(incl) => {
                         for t in self.sorter.sort(incl) {
-                            self.producer.enqueue(t).await.map_err(|_| TimeboostError::ChannelClosed)?
+                            if let Err(err) = self.producer.enqueue(t).await {
+                                error!(node = %self.label, %err, "failed to enqueue transaction");
+                                return Err(TimeboostError::ChannelClosed);
+                            }
                         }
                         if self.decrypter.has_capacity() {
                             let Some(ilist) = pending.take() else {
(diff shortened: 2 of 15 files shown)


timeboost-sequencer/src/lib.rs
```diff
@@ -25,12 +28,15 @@ use tracing::{error, info, warn};
...
 use decrypt::{DecryptError, Decrypter};
 use include::Includer;
+use multiplex::Multiplex;
 use queue::BundleQueue;
 use sort::Sorter;
 
 type Result<T> = std::result::Result<T, TimeboostError>;
 type Candidates = VecDeque<(RoundNumber, Vec<CandidateList>)>;
 
+const MAX_SIZE: usize = 100;
```
twittner, line 38: This name is fairly generic and deserves at least a comment I think.

timeboost-sequencer/src/lib.rs
```diff
@@ -344,15 +376,9 @@ impl Task {
                 candidates.push((round, lists))
             }
             while let Some(action) = actions.pop_front() {
-                match action {
-                    Action::Deliver(_) => {
-                        actions.push_front(action);
-                        break;
-                    }
-                    Action::Gc(r) => {
```
twittner, line 352: Why is this removed?

timeboost-sequencer/src/multiplex.rs
```diff
@@ -0,0 +1,249 @@
...
+            block_tx,
+            rx,
+            gc_rx,
+        }
+    }
+    pub async fn go(mut self) {
+        loop {
+            tokio::select! {
+                Some(message) = self.rx.recv() => {
+                    let data = match serialize(&message) {
```
twittner, line 161: It would be nice if the multiplexer could work with `Data`. If serialisation fails here, e.g. due to length violation, the sender would have no way to detect this.

timeboost-sequencer/src/multiplex.rs
```diff
@@ -0,0 +1,249 @@
...
+/// Serialize a given data type into `Bytes`
+fn serialize<T: Serialize>(d: &T) -> Result<Data> {
+    let mut b = BytesMut::new().writer();
+    bincode::serde::encode_into_std_write(d, &mut b, bincode::config::standard())?;
+    Ok(b.into_inner().try_into()?)
+}
+
+/// Deserialize from `Bytes` into a given data type.
+fn deserialize<T: for<'de> serde::Deserialize<'de>>(d: &bytes::Bytes) -> Result<T> {
+    bincode::serde::decode_from_slice(d, bincode::config::standard())
```
twittner, line 233: This should set a byte limit on the bincode config.

timeboost-sequencer/src/multiplex.rs
```diff
@@ -0,0 +1,249 @@
...
+                            if let Err(e) = self.block_tx.send((public_key, envelope)).await {
+                                error!("failed to send block message: {}", e);
+                                return;
+                            }
+                        }
+                    }
+                }
+                Some(bucket) = self.gc_rx.recv() => {
+                    trace!(node = %self.label, "received gc signal for bucket: {:?}", bucket);
+                    self.net.gc(bucket);
```
twittner, line 213: NB that `Overlay::gc` removes all buckets less than the given one.

timeboost-sequencer/src/lib.rs
akonring: Good point. Fixed.

timeboost-sequencer/src/lib.rs
```diff
@@ -213,9 +229,9 @@ impl Sequencer {
         self.bundles.add_bundles(it)
     }
 
-    pub async fn next_transaction(&mut self) -> Result<Transaction> {
+    pub async fn next_block(&mut self) -> Result<CertifiedBlock> {
```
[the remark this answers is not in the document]
akonring: Agreed, that approach makes sense. Given the scope of this PR, I’ll leave this as a follow-up migration if that works.

timeboost-sequencer/src/lib.rs
akonring: Thanks. Fixed.

timeboost-sequencer/src/multiplex.rs
```diff
@@ -0,0 +1,249 @@
...
+type DecryptReceiver = Receiver<(PublicKey, ShareInfo)>;
+type BlockReceiver = Receiver<(PublicKey, BlockInfo)>;
+
+struct BlockBucket(u64);
+
+impl From<u64> for BlockBucket {
+    fn from(value: u64) -> Self {
+        let mut bucket = value;
+        bucket ^= 1 << 63;
+        Self(bucket)
```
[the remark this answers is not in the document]
akonring: fixed.

multisig/src/envelope.rs
```diff
@@ -71,6 +71,15 @@ impl<D: Committable, S> Envelope<D, S> {
...
 
+impl<D: Committable> From<Envelope<D, Validated>> for Envelope<D, Unchecked> {
+    fn from(value: Envelope<D, Validated>) -> Self {
+        Envelope {
+            signed: value.signed,
+            _marker: PhantomData,
+        }
+    }
+}
+
```
twittner, line 82:
```suggestion
```

I think this is not needed (see the comment on `BlockInfo`).

timeboost-sequencer/src/multiplex.rs
akonring: Yes, this should be fixed in the refactored version. Also, Decryption and Block Production do not share the same outbound channel anymore (there is a dedicated channel to each phase now).

timeboost-sequencer/src/multiplex.rs
akonring: Thanks. Fixed.

timeboost-types/src/block.rs
```diff
@@ -0,0 +1,198 @@
...
+            .fixed_size_field("block-hash", &self.0)
+            .finalize()
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct BlockInfo {
+    num: BlockNumber,
+    envelope: Envelope<BlockHash, Unchecked>,
+}
```
twittner, line 153:
```suggestion
pub struct BlockInfo<S> {
    num: BlockNumber,
    envelope: Envelope<BlockHash, S>,
}
```

Making this generic over the validation state of the `Envelope` allows you to deserialise an `Envelope<BlockHash, Unchecked>` and to serialise an `Envelope<BlockHash, Validated>` or `Envelope<BlockHash, Unchecked>`.

timeboost-sequencer/src/multiplex.rs
```diff
@@ -0,0 +1,303 @@
...
+const DECRYPT_TAG: u8 = 0xDE;
+const BLOCK_TAG: u8 = 0xB0;
+
+type Result<T> = std::result::Result<T, MultiplexError>;
+
+struct BlockBucket(u64);
+
+impl From<u64> for BlockBucket {
+    fn from(value: u64) -> Self {
+        Self(value ^ (1 << 63))
```
twittner, line 25: One question about this -- Is the intention to set the most significant bit for block buckets or to toggle it? I would expect the former, and if so, should this not be `value | (1 << 63)`?

akonring: Fixed. Both should work, imo. The underlying assumption is that `value` is less than $$2^{63}$$ so it shouldn't matter if or or xor is applied I think.

twittner (approved):

PR #329, opened 2025-04-24
Avoid non-trivial address recovery in tx compare

akonring:
While profiling Timeboost it became evident that computing address recovery consumes 20-40% of the CPU time (with 100 TPS). This PR fixes this and adds an additional SortedTransaction which is the output of the Sorting phase. For now just containing signer address, timestamp, transactions.

### DIFF
--- a/timeboost-sequencer/src/sort.rs
+++ b/timeboost-sequencer/src/sort.rs
@@ -1,7 +1,7 @@
 use std::cmp::Ordering;
 
 use ssz::decode_list_of_variable_length_items as ssz_decode;
-use timeboost_types::{Bytes, InclusionList, Timestamp, Transaction};
+use timeboost_types::{Bytes, InclusionList, SortedTransaction, Transaction};
 use tracing::warn;
 
 const MAX_BUNDLE_TXS: usize = 1024;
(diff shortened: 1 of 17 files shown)


timeboost-sequencer/src/sort.rs
```diff
@@ -37,10 +37,14 @@ impl Sorter {
                         }
                         match Transaction::decode(&t) {
                             Ok(tx) => {
+                                let Ok(signer) = tx.tx().recover_signer() else {
```
twittner, line 40: It seems that transactions where signer recovery fails are now dropped, whereas before only the comparison would be affected. Requiring signer recovery is stricter and more straightforward, however would it not make more sense to reject transactions *on submission* when the recovery fails? Why even including them in candidate lists?

timeboost-types/src/bundle.rs
```diff
@@ -377,11 +377,50 @@ impl Transaction {
     }
 }
 
+pub struct SortedTransaction {
```
twittner, line 380: Maybe we can include recovered addresses directly in `Transaction` and drop this `SortedTransaction` type?

timeboost-sequencer/src/sort.rs
twittner: I guess encrypted transactions get in the way of that approach.

jparr721 (approved):

PR #337, opened 2025-04-30
Use a custom bounded channel for peers.

twittner:
Since layers above `Network` have to cope with message loss they usually perform retries (cf. `Overlay`), sending the same message until they received acknowledgements from peers. If the peer is slow, this may cause multiple identical copies of the message to appear in the queue and subsequently sent over the network. This commit keeps only a single message (as identified by its ID) in the channel. Either this message is eventually sent out or the channel is dropped when it overflows, or a timeout occurs (just like currently).

### DIFF
--- a/cliquenet/src/chan.rs
+++ b/cliquenet/src/chan.rs
@@ -1,0 +1,100 @@
+//! A channel implementation that keeps only a single copy of an item, as
+//! identified by its Id.
+
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+use nohash_hasher::IntSet;
+use parking_lot::Mutex;
+use tokio::sync::Notify;
+use tokio::sync::mpsc::error::TrySendError;
+
+use crate::Id;
+
+#[derive(Debug)]
+pub struct Sender<T>(Arc<Chan<T>>);
+
+#[derive(Debug)]
+pub struct Receiver<T>(Arc<Chan<T>>);
+
+#[derive(Debug)]
+struct Chan<T> {
+    /// Channel capacity.
+    cap: usize,
+    /// Notifier for receivers that are waiting for items.
+    sig: Notify,
+    /// The items currently in flight.
+    buf: Mutex<Buf<T>>,
+}
+
+#[derive(Debug)]
+struct Buf<T> {
+    /// Ordered queue of items.
+    xs: VecDeque<(Option<Id>, T)>,
+    /// The set of Ids in the queue.
+    ids: IntSet<Id>,
+}
+
+impl<T> Clone for Sender<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
+
+pub fn channel<T>(cap: usize) -> (Sender<T>, Receiver<T>) {
+    let chan = Arc::new(Chan {
+        cap,
+        sig: Notify::new(),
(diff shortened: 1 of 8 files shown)


jparr721 (approved):

cliquenet/src/chan.rs
```diff
@@ -0,0 +1,100 @@
...
+        }),
+    });
+    (Sender(chan.clone()), Receiver(chan))
+}
+
+impl<T> Sender<T> {
+    pub fn try_send(&self, id: Option<Id>, val: T) -> Result<(), TrySendError<T>> {
+        if let Some(id) = id {
+            let mut buf = self.0.buf.lock();
+            if buf.ids.contains(&id) {
```
lukeiannucci, line 60: I know you explained this to me once, but why can the buffer already contain this id?

cliquenet/src/chan.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Notifier for receivers that are waiting for items.
+    sig: Notify,
+    /// The items currently in flight.
+    buf: Mutex<Buf<T>>,
+}
+
+#[derive(Debug)]
+struct Buf<T> {
+    /// Ordered queue of items.
+    xs: VecDeque<(Option<Id>, T)>,
```
lukeiannucci, line 33: not following the naming here :laughing:

cliquenet/src/chan.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn try_send(&self, id: Option<Id>, val: T) -> Result<(), TrySendError<T>> {
+        if let Some(id) = id {
+            let mut buf = self.0.buf.lock();
+            if buf.ids.contains(&id) {
+                return Ok(());
+            }
+            if buf.xs.len() == self.0.cap {
+                return Err(TrySendError::Full(val));
+            }
+            buf.xs.push_back((Some(id), val));
```
lukeiannucci, line 66: why do we need `xs` and `ids` doesnt `xs` have the id? could we not change the data structure for `xs`? or is this too complex for what it needs to do, just push and pop?

cliquenet/src/net.rs
```diff
@@ -527,7 +529,7 @@ where
                 _ = self.ping_interval.tick() => {
                     let now = Timestamp::now();
                     for task in self.active.values() {
-                        let _ = task.tx.try_send(Message::Ping(now));
+                        let _ = task.tx.try_send(None, Message::Ping(now));
```
lukeiannucci, line 532: so is the idea here that if it is a ping pong message we always want to include? thats why we set the id to none?

lukeiannucci (approved):
this looks good to me, just left some simple questions for my understanding as i ramp back up

cliquenet/src/chan.rs
twittner: We have `xs : VecDeque<(Option<Id>, T)>`. Membership testing of an `y : Id` would be $$O(n)$$, hence the separate hash set containing only the `Id`s of the channel, which allows us to do $$O(1)$$ checks.

cliquenet/src/net.rs
twittner: Yes, any message that is sent without an `Id` is not subject to deduplication.

PR #338, opened 2025-04-30
Share `Multiplex`.

twittner:
Rather than handing out MPSC senders, have decrypt and block producer share the `Multiplex`er directly. Internally, unify the various message types sent to the worker.


jparr721 (approved):

twittner:
Closing in preference to #339.

PR #340, opened 2025-05-02
Improve threshold encryption

### LINKED ISSUE #336: comments on Shoup-Gennaro threshold decryption scheme
I read through our current implementation, and have the following comments for future fix and/or improvements:

## Security related 

- Add [`Zeroize` ](https://docs.rs/zeroize/latest/zeroize/index.html)to clean up memory on drop with sensitive info/keys, including
  - https://github.com/EspressoSystems/timeboost/blob/d786e8458b5595bf139c21445c0acd9143c7a181/timeboost-crypto/src/lib.rs#L107
  - ❗ this is more critical as `alpha_0` is essentially a trapdoor for the entire scheme: https://github.com/EspressoSystems/timeboost/blob/d786e8458b5595bf139c21445c0acd9143c7a181/timeboost-crypto/src/sg_encryption.rs#L67 

## Improvements 

- `struct Plaintext` should contain both raw message and associated data, see [here](https://docs.rs/aead/0.5.2/aead/struct.Payload.html), this would allow appropriate AEAD interface, sometimes people want some associated data (such as header or some metadata) with the cipertext
- during `setup()` the degree should be `threshold - 1` 
  https://github.com/EspressoSystems/timeboost/blob/d786e8458b5595bf139c21445c0acd9143c7a181/timeboost-crypto/src/sg_encryption.rs#L60
  such that later in recovery (decryption) we don't break the original intended meaning of "threshold" by hacking the boundary condition: 
  https://github.com/EspressoSystems/timeboost/blob/d786e8458b5595bf139c21445c0acd9143c7a181/timeboost-crypto/src/sg_encryption.rs#L175
-  use [FFT ](https://docs.rs/ark-poly/latest/ark_poly/domain/trait.EvaluationDomain.html#method.fft)instead of evaluate one by one:
  https://github.com/EspressoSystems/timeboost/blob/d786e8458b5595bf139c21445c0acd9143c7a181/timeboost-crypto/src/sg_encryption.rs#L68-L73

## Miscanllenous 

- Consider other CCA secure symmetric AEAD candidates, even though AES-GCM has great intel AES-NI instruction support, ChaCha20-Poly1305 is more [side-channel resilient  ](https://datatracker.ietf.org/doc/html/rfc8439), or the candidates from the winner of [CAESAR competition](https://competitions.cr.yp.to/caesar-submissions.html).

## Question

- the main purpose of [`SerdeAs`](https://github.com/EspressoSystems/timeboost/blob/d786e8458b5595bf139c21445c0acd9143c7a181/timeboost-crypto/src/lib.rs#L321) is to override the default arkwork behavior of uncompressed serialization? and also packed representation for vectors (since this will skip the "append vector length" canonical serialization behavior): 
https://github.com/EspressoSystems/timeboost/blob/d786e8458b5595bf139c21445c0acd9143c7a181/timeboost-crypto/src/lib.rs#L92

alxiong:
Closes #336 

### This PR:

Address all security and performance issues listed in the github issues (linked above).

Here I want to particularly talk about ddad67f79ea349909caea949330a2cd8814f81df, I did the following two performance optimization: 
1. use [Barycentric form](https://en.wikipedia.org/wiki/Lagrange_polynomial#Barycentric_form) to compute lagrange coefficient for efficiently than the naive iteration
2. use arkworks' `::msm()` when interpolating in the exponent, (instead of naively scalar mul them and add together) 

I ran benchmark locally, when the number of nodes is small (like 5) the improvement is neglibile as expected, only 2%. 
when we are scaling to larger network, such as 20 nodes, this commit alone improves the `combine()` speed by 4~8%. which is not significant but also not too bad. 

Here are some sampled benchmark instances: 
```
SG01_Bn254G1_combine_100KB/20
                        time:   [27.369 ms 27.372 ms 27.376 ms]
                        thrpt:  [3.5673 MiB/s 3.5677 MiB/s 3.5681 MiB/s]
                 change:
                        time:   [-4.4494% -4.1859% -3.9488%] (p = 0.00 < 0.05)
                        thrpt:  [+4.1111% +4.3688% +4.6566%]
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)

SG01_Bn254G1_combine_200KB/20
                        time:   [28.676 ms 28.679 ms 28.683 ms]
                        thrpt:  [6.8094 MiB/s 6.8102 MiB/s 6.8109 MiB/s]
                 change:
                        time:   [-9.9294% -6.1109% -3.4077%] (p = 0.00 < 0.05)
                        thrpt:  [+3.5279% +6.5086% +11.024%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)

SG01_Bn254G1_combine_1024KB/20
                        time:   [34.690 ms 34.730 ms 34.799 ms]
                        thrpt:  [28.736 MiB/s 28.794 MiB/s 28.827 MiB/s]
                 change:
                        time:   [-7.9358% -5.5677% -3.6814%] (p = 0.00 < 0.05)
                        thrpt:  [+3.8221% +5.8960% +8.6198%]
                        Performance has improved.
```

### This PR does not:

The miscanllenous and questions section of the issue description is ignored


Cargo.toml
```diff
@@ -90,3 +90,4 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter", "json"] }
 turmoil = "0.6.4"
 vbs = "0.1"
+zeroize = { version = "^1.8", features = ["zeroize_derive"] }
```
twittner, line 93:
```suggestion
zeroize = { version = "1.8", features = ["zeroize_derive"] }
```
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

twittner (approved):
Thanks Alex! :+1:

timeboost-crypto/src/sg_encryption.rs
```diff
@@ -170,9 +178,10 @@ where
         comb_key: &Self::CombKey,
         dec_shares: Vec<&Self::DecShare>,
         ciphertext: &Self::Ciphertext,
+        aad: &Self::AssociatedData,
     ) -> Result<Self::Plaintext, ThresholdEncError> {
         let committee_size: usize = committee.size.get();
-        let threshold = committee.threshold().get() + 1;
+        let threshold = committee.threshold().get();
```
alxiong, line 184: after seeing how sequencer is using this code, i realize that my change (introduced in [9870e7f](https://github.com/EspressoSystems/timeboost/pull/340/commits/9870e7f062d17014ed25b1d8aaff024ad307f1e9)) is probably wrong.

when i made the change, i was looking at the scheme impl in isolation, and assume we can take the "threshold" on its standard `(t, N)`-threshold definition. but here we need to set a value congruent with consensus's threshold. so `t+1` makes more sense. 

I will try to explain more in an issue and fix my own mess.

PR #341, opened 2025-05-05
Create timeboost-builder crate.

twittner:
Move the block production into its own crate (`timeboost-builder`). The `Sequencer` of `timeboost-sequencer` now again only produces an ordered sequence of transactions. The timeboost crate ties the two together, enqueuing the transactions in the builder.

Also make less implicit assumptions about ports, but have the various network addresses be part of the configuration.

The transaction_order test has been moved into the tests crate and is accompanied by a block_order test, with both sharing the same setup logic.

Finally, `NetworkMetrics` requires a label now, to disambiguate metrics per network.

### DIFF
--- a/test-configs/local.json
+++ b/test-configs/local.json
@@ -1,121 +1,161 @@
 {
     "keyset": [
         {
-            "url": "127.0.0.1:8000",
+            "sailfish_url": "127.0.0.1:8000",
+            "decrypt_url": "127.0.0.1:9000",
+            "producer_url": "127.0.0.1:10000",
             "pubkey": "HCVAkbpaXjt4NqHem1Qz2mHqKRbGFCHUf4VMbiB7jAYE",
             "sig_pk": "24f9BtAxuZziE4BWMYA6FvyBuedxU9SVsgsoVcyw3aEWagH8eXsV6zi2jLnSvRVjpZkf79HDJNicXSF6FpRWkCXg",
             "dec_pk": "jMbTDiLo8tgyERv92mGrCAe1s3KnnnyqhQeSYte6vUhZy1"
         },
         {
-            "url": "127.0.0.1:8001",
+            "sailfish_url": "127.0.0.1:8001",
+            "decrypt_url": "127.0.0.1:9001",
+            "producer_url": "127.0.0.1:10001",
             "pubkey": "6v8B9RK6Noa6SX8rE6RXK972v6KFfVaLjC5sbCT9VoJn",
             "sig_pk": "2gtHurFq5yeJ8HGD5mHUPqniHbpEE83ELLpPqhxEvKhPJFcjMnUwdH2YsdhngMmQTqHo9B1Qna6uM13ug2Pir97k",
             "dec_pk": "jysmvvvwSHu872gmxkejPP8RxUDpSpKChnkPMVeXyRibwN"
         },
         {
-            "url": "127.0.0.1:8002",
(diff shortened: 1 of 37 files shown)


test-configs/local.json
```diff
@@ -1,121 +1,161 @@
...
+            "sailfish_url": "127.0.0.1:8018",
+            "decrypt_url": "127.0.0.1:10018",
+            "producer_url": "127.0.0.1:11018",
             "pubkey": "5qP4wyDetRcVzQZs3qcxyafxXYowBTdcSVQJbHXcZsji",
             "sig_pk": "kCewFsRMYVPQ59rKCQUtEqjnrezJpZH82GqDYtwCZMWD35fgJhFV535YXrNzE1hLXgfXdD5o1DHsnvHwNwweLX6",
             "dec_pk": "kEaesywh1xhXjUe7fzYPnNHJo9i4kFwdN3paA569MpbuLu"
         },
         {
-            "url": "127.0.0.1:8019",
+            "sailfish_url": "127.0.0.1:8019",
```
lukeiannucci, line 156: in `docker-compose.yml` do we need to expose these ports now too?

timeboost-builder/src/produce.rs
```diff
@@ -33,7 +31,7 @@ pub struct BlockProducer {
     /// Keypair of the node.
     label: Keypair,
     /// Incoming transactions
-    queue: VecDeque<(Timestamp, Transaction)>,
+    queue: VecDeque<Transaction>,
```
lukeiannucci, line 34: why is the timestamp not needed?

timeboost-builder/src/produce.rs
```diff
@@ -121,21 +144,27 @@ impl BlockProducer {
...
-                        block = %num,
-                        next  = %block_info.0,
-                        "received future certified block",
-                    );
-                    self.blocks.insert(block_info, status);
-                }
+            if let Some(cb) = self.first_certified() {
+                return Ok(cb);
+            } else {
+                debug!(node = %self.label.public_key(), %num, "received future certified block");
```
lukeiannucci, line 150: is this really common? or do we maybe want to increase this to `info`?

lukeiannucci (approved):
This looks good to me. However, I am still ramping back up with everything.

timeboost-builder/src/produce.rs
twittner: The `Timestamp` is now part of `Transaction`.

timeboost-builder/src/produce.rs
twittner: Receiving block information out of order is normal and can happen any time.

test-configs/local.json
lukeiannucci: i ran did this commit:
https://github.com/EspressoSystems/timeboost/pull/341/commits/c8eff1b5c481bc68c8cd0ef9e0601a5053553ea3

it at least starts properly in docker now. however, i see this panic:
```text
2025-05-06 16:28:00 thread 'tokio-runtime-worker' panicked at timeboost-crypto/src/sg_encryption.rs:210:32:
2025-05-06 16:28:00 index out of bounds: the len is 20 but the index is 213
2025-05-06 16:28:00 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-05-06 16:28:00 Error: timeboost shutdown unexpectedly
```

PR #342, opened 2025-05-07
SG01 refactor 

alxiong:
This is Part 1 of the Decryption phase refactor effort: https://app.asana.com/1/1208976916964769/project/1209394409339229/task/1209978846980044

### This PR:

- rename fuzzy `keyset.threshold()` to explicit `.one_honest_threshold()`
- add a missing ciphertext validity check during `combine()` (Step1.1 in BonehShoup Sec 22.3.3.2)
- extend error type to support `ThresholdEncError::FaultySubset` which outputs a blameable faulty subset for downstream users/callers to deal with those faulty decryption shares. 
  - unit tests updated to test this functionality

### DIFF
--- a/timeboost-crypto/src/sg_encryption.rs
+++ b/timeboost-crypto/src/sg_encryption.rs
@@ -107,6 +107,7 @@ where
         let w = pub_key.key * beta;
 
         // hash to symmetric key `k`
+        // TODO: (alex) consider moving kid as part of `aad` instead
         let key = hash_to_key::<C, H>(v, w, (*kid).into())
             .map_err(|e| ThresholdEncError::Internal(anyhow!("Hash to key failed: {:?}", e)))?;
         let k = GenericArray::from_slice(&key);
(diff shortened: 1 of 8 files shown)


timeboost-crypto/src/sg_encryption.rs
```diff
@@ -202,21 +202,19 @@ where
...
-        // Verify DLEQ proofs
-        let valid_shares: Vec<_> = dec_shares
-            .iter()
-            .filter_map(|share| {
+        // Verify DLEQ proofs to ensure correctness of the decryption share w_i
+        // keeping a list of faulty shares (their node indices) to blame
+        let (valid_shares, faulty_shares): (Vec<&Self::DecShare>, Vec<_>) =
+            dec_shares.iter().partition(|share| {
                 let (w, phi) = (share.w, share.phi.clone());
                 let u = pk_comb[share.index as usize];
```
twittner, line 210: Not related to this PR, but maybe worth including: Using `Vec::get` instead of the index operator here is probably a good idea, given that these values are passed in as arguments.

twittner (approved):
:+1:

timeboost-crypto/src/sg_encryption.rs
alxiong: sounds reasonable, but this will spoil the `partition()` structure. I'll update the PR

alxiong: fixed in [68ee2df](https://github.com/EspressoSystems/timeboost/pull/342/commits/68ee2df52f3c78a10b6f8ca5ceccb6c0936434ff)

PR #343, opened 2025-05-07
Return transactions in batches from sequencer.

twittner:
Instead of enqueuing the individual transactions between sequencer worker task and then consuming them one by one, put the whole available batch into the channel and also return that to the caller. Clients can still flatten the batches into one sequence of transactions but potentially they can also process the batches in a more performant way.


lukeiannucci (approved):
this makes sense to me! curious does this help performance batching the transactions then sending the batch across the channel?

PR #344, opened 2025-05-07
Use Proper Docker Config for Decryption Phase

lukeiannucci:
Closes #<ISSUE_NUMBER>

### This PR:

We noticed that during the decryption phase we were indexing incorrectly resulting in code panicking. Alex got a pr in to log the error instead of a panic, see discussion here:
https://github.com/EspressoSystems/timeboost/pull/342#discussion_r2077374873

We only noticed this when running in docker as local integration worked fine.
A simple fix for now is to update the `docker.json` to be the same as `local.json` as we noticed the public keys for each node between the files were the same. 
### This PR does not:

We still have a lot of question marks around how to properly generate and update the configuration. We need to understand this flow end to end.

### Things tested

On my local docker builds I added logging to ensure we decrypt the data properly at the end of `combine()` in `sg_encryption.rs`  and ensure we never see this error in my docker nodes logs:
```rust
ThresholdEncError::Argument(format!("CombKey missing idx: {}", share.index))
```

### DIFF
--- a/test-configs/docker.json
+++ b/test-configs/docker.json
@@ -4,125 +4,165 @@
             "sailfish_url": "172.20.0.2:8000",
             "decrypt_url": "172.20.0.2:8001",
             "producer_url": "172.20.0.2:8002",
-            "pubkey": "HCVAkbpaXjt4NqHem1Qz2mHqKRbGFCHUf4VMbiB7jAYE"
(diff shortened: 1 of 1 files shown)


alxiong (approved):
let's merge this first.

PR #353, opened 2025-05-20
Retry again in RBC worker.

twittner:
In case a get request is either not answered or replied to with an invalid response a different voter needs to be asked for the message. Since the `Overlay` network does not know about this logic, the RBC worker itself needs to periodically check messages with `Status::Requested` and retry.

Also fixes some clippy warnings.

### DIFF
--- a/sailfish-rbc/src/abraham/worker.rs
+++ b/sailfish-rbc/src/abraham/worker.rs
@@ -15,7 +15,7 @@ use multisig::{Unchecked, Validated};
 use sailfish_types::{Evidence, Message, RoundNumber, Vertex};
 use serde::{Serialize, de::DeserializeOwned};
 use tokio::sync::mpsc;
-use tokio::time::Instant;
+use tokio::time::{self, Duration, Instant, Interval};
 use tracing::{debug, error, trace, warn};
 
 use crate::RbcError;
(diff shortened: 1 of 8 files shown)


sailfish-rbc/src/abraham/worker.rs
```diff
@@ -837,6 +880,36 @@ impl<T: Clone + Committable + Serialize + DeserializeOwned> Worker<T> {
...
+                );
+                let timeout = [1, 3, 6, 10, 15].get(tracker.retries).copied().unwrap_or(30);
+                if tracker.timestamp.elapsed() < Duration::from_secs(timeout) {
+                    continue;
+                }
+                debug!(node = %self.key, %digest, "requesting message again");
+                let m = Protocol::<'_, T, Validated>::GetRequest(*digest);
+                let b = serialize(&m)?;
+                let c = digest.commit();
+                let s = tracker.choose_voter(&c).expect("status = requested => voter");
```
lukeiannucci, line 903: i am a bit confused why we always choose the same voter. do we only send this get request to one node in the committe?

twittner: https://github.com/EspressoSystems/timeboost/blob/845c7a8feecf67cfac176df6a2dde24a71e026e1/sailfish-rbc/src/abraham/worker.rs#L133-L136
`choose_voter` selects a random voter.

lukeiannucci: probably should have looked into the method. makes sense!

PR #354, opened 2025-05-20
Update dependencies

twittner:


sailfish-consensus/src/lib.rs
```diff
@@ -582,6 +582,7 @@ where
     /// If all edges of the vertex point to other vertices in the DAG we add the
     /// vertex to the DAG. If we also have more than 2f vertices for the given
     /// round, we can try to commit the leader vertex of a round.
+    #[allow(clippy::result_large_err)]
     fn try_to_add_to_dag(&mut self, v: Vertex<T>) -> Result<Vec<Action<T>>, Vertex<T>> {
```
lukeiannucci, line 586: why is this needed now?

lukeiannucci (approved):

sailfish-consensus/src/lib.rs
twittner: Clippy in Rust 1.87 was complaining.

PR #356, opened 2025-05-23
Switch signing keys from Ed25519 to Secp256k1.

twittner:
This is mostly for compatibility reasons. This commit also changes the configuration to include an explicit keypair for X25519 key exchange which is used for establishing secure channels via Noise in the network layer.


tests/src/tests/network/network_tests.rs
```diff
@@ -279,18 +279,18 @@ where
...
-                .keypairs
+                .sign_keypairs
                 .iter()
                 .map(|kpr| {
                     let node_public_key = kpr.public_key();
                     TestCondition::new(format!("Vertex from {}", k.public_key()), move |msg, _a| {
                         if let Some(Message::Vertex(v)) = msg {
                             let r = **v.data().round().data();
-                            if v.data().evidence().is_timeout() && r != 3 {
+                            if v.data().evidence().is_timeout() && r != 6 {
```
lukeiannucci, line 293: help me understand how this test is working because i am a bit confused, the test tells the node to come online at round 4. is it because the dag for the node is not yet ready from catchup?

lukeiannucci (approved):
nice work, this looks good to me just some questions

multisig/src/lib.rs
```diff
@@ -47,104 +45,102 @@ impl From<KeyId> for u64 {
...
 
     /// Sign data with our ed25519 secret key.
     pub fn sign(&self, data: &[u8], deterministic: bool) -> Signature {
-        Signature {
-            sig: self
-                .pair
-                .sk
-                .sign(data, (!deterministic).then(ed25519::Noise::generate)),
-        }
+        self.sk.sign(data, deterministic)
```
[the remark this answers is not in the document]
twittner: Thanks, I forgot to remove it.

tests/src/tests/network/network_tests.rs
twittner: The timeout depends on when the node with ID 4 is leader of a round and the leader selection is dependent on the ordering of public keys. Due to the change from Ed25519 to Secp256k1 this order has changed and Node 4 is leader at a later round, resulting in a later timeout.

timeboost-sequencer/src/decrypt.rs
```diff
@@ -714,11 +714,18 @@ mod tests {
 
     async fn setup(keyset: Keyset) -> Vec<Decrypter> {
         let signature_private_keys = [
-            "24f9BtAxuZziE4BWMYA6FvyBuedxU9SVsgsoVcyw3aEWagH8eXsV6zi2jLnSvRVjpZkf79HDJNicXSF6FpRWkCXg",
-            "2gtHurFq5yeJ8HGD5mHUPqniHbpEE83ELLpPqhxEvKhPJFcjMnUwdH2YsdhngMmQTqHo9B1Qna6uM13ug2Pir97k",
-            "4Y7yyg11MBYJaeD2UWCh5cto8wizJoUCqFm7YjMbY3hXyWSWVPgi7y1D9e7D78a1NcWdE4k59D6vK9f6eCpzVkbQ",
-            "zrjZDknq9nPhiBXKeG2PeotwxAayYkv2UFmxc2UsCGHsdu5vCsjqxn8ko2Rh2fWts76u6eCJYjDgKEaveutVhjW",
-            "jW98dJM94zuvhRCA1bGLiPjakePTc1CYPP2V5iCswfayZiYujGYdSoE1MYDa61dHCyzPdEvGNBDmnFHS6jf83Km",
+            "4aZN8RbcnmVkWvhX9D78qcu62eFc7JuWvy2Yf8SLrjey",
```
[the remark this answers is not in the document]
twittner: I don't know. The test already existed.

tests/src/tests/network/network_tests.rs
lukeiannucci: Yes I get this, but the node should come online at round 4. And timeout is happening on round 6

twittner: If node 4 is the first leader it times out in round 0. Since 0 is the genesis round, there is an exception regarding the required evidence. Node 4 is leader again in round 5 where it times out because it only sees messages from round 5 which it can not at to its DAG. This timeout evidence is then visible in the next round which is round 6.

lukeiannucci: thanks, for the details. this makes sense and i suspected this is what was happening. just wanted to make sure i understood it

multisig/src/lib.rs
```diff
@@ -47,104 +45,102 @@ impl From<KeyId> for u64 {
...
-        PublicKey {
-            key: self.key.public_key(),
-        }
+        let pk = self.key.public_key(secp256k1::SECP256K1);
+        PublicKey { key: pk }
     }
 
-    pub fn as_bytes(&self) -> [u8; 64] {
-        *self.key
+    pub fn as_bytes(&self) -> [u8; 32] {
```
alxiong, line 142: there are many cases of `as_bytes` and `to_bytes` i don't know the convention we are following.

multisig/src/util.rs
```diff
@@ -1,40 +1,107 @@
...
+    } else {
+        let a = ByteArray::<32>::deserialize(d)?;
+        secp256k1::SecretKey::from_byte_array(a.into_array()).map_err(de::Error::custom)
+    }
 }
 
-pub(crate) fn decode_sig<'de, D>(d: D) -> Result<ed25519::Signature, D::Error>
+pub(crate) fn encode_secp256k1_pk<S>(k: &secp256k1::PublicKey, s: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
```
alxiong, line 82: may I ask why do we need to duplicate this impl, can we just use the generic 

```
fn encode<S, T, const N: usize>(d: &T, s: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
    T: Serialize,
```

multisig/src/util.rs
```diff
@@ -1,40 +1,107 @@
...
+    if d.is_human_readable() {
+        let s = String::deserialize(d)?;
+        let mut a = [0; 32];
+        let n = bs58::decode(&s).onto(&mut a).map_err(de::Error::custom)?;
+        x25519::SecretKey::from_slice(&a[..n]).map_err(de::Error::custom)
+    } else {
+        let a = ByteArray::<32>::deserialize(d)?;
+        x25519::SecretKey::from_slice(&a[..]).map_err(de::Error::custom)
+    }
+}
```
alxiong, line 49: i'm unsure about our serde strategy, esp why do we have to convert these objects into ByteArray first? 

the "per-object" serde functions also kinda of baffles me.

multisig/src/x25519.rs
```diff
@@ -43,6 +73,11 @@ impl PublicKey {
 }
 
 impl SecretKey {
+    pub fn public_key(&self) -> PublicKey {
+        let key = self.key.recover_public_key().expect("valid public key");
+        PublicKey { key }
+    }
+
```
alxiong, line 80: nit: implemented as `From<SecretKey> for PublicKey`? (doesn't really matter, just a personal preference i guess)

multisig/src/lib.rs
twittner: We should follow these https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

multisig/src/util.rs
twittner: This is a good point. I think for the non-human readable format this is not required.

twittner: Actually the `Serialize` and `Deserialize` implementations of the various `secp256k1` types use the regular sequential forms of serde and here, similar to the implementations for the x25519 keys, we use the optimised `serde_bytes` form. This results for example in a single `Serializer::serialize_bytes` call instead of serializing the key as a tuple and then each element.

PR #357, opened 2025-05-23
Use IETF standardized hash to curve

alxiong:
Closes https://app.asana.com/1/1208976916964769/project/1209394409339229/task/1209993325108067?focus=true

### This PR:

- Switch to IETF compliant HashToCurve, and currently only support Bls12-381 curve for threshold decryption (see Asana ticket comment I left: https://app.asana.com/1/1208976916964769/task/1209993325108067/comment/1210279951025516?focus=true)
- Regenerate all keys with `just keygen --seed 0000000000000000000000000000000000000000000000000000000000000000 --num 13 --out test-keys` and manually update `test-configs/*.json` files
  - I cut down the keyset size from 20 active members to 13. Happy to set to 20 if we want in the future. This is easily configurable and updatable.


alxiong:
@twittner the config files in this PR still needs to be correctly updated. But I'd like to use your script in #358 after its merged.

For now, the `just ci_local` and `cargo test -p timeboost-crypto` are all passing. so feel free to review the `crypto` part of the code which should be straightforward. 

I will make the CI pass later.

justfile
```diff
@@ -81,6 +81,19 @@ run *ARGS:
...
+  cargo run --bin mkconfig -- -n {{NUM_NODES}} --sailfish-base-addr "127.0.0.1:8000" --decrypt-base-addr "127.0.0.1:10000" --producer-base-addr "127.0.0.1:11000" --mode "increment-port" {{ARGS}} | jq
+
+mkconfig_docker NUM_NODES *ARGS:
+  cargo run --bin mkconfig -- -n {{NUM_NODES}} --sailfish-base-addr "172.20.0.2:8000" --decrypt-base-addr "172.20.0.2:8001" --producer-base-addr "172.20.0.2:8002" --mode "increment-address" {{ARGS}} | jq
+
+mkconfig_cloud_single:
+  cargo run --bin mkconfig -- -n 5 \
+  --sailfish-base-addr "timeboost-lb-50d346419d44f480.elb.us-east-2.amazonaws.com:8000" \
+  --decrypt-base-addr "timeboost-lb-50d346419d44f480.elb.us-east-2.amazonaws.com:10000" \
+  --producer-base-addr "timeboost-lb-50d346419d44f480.elb.us-east-2.amazonaws.com:11000" \
```
twittner, line 94: These are very specific. Not sure if they should be in here.

timeboost/src/binaries/mkconfig.rs
```diff
@@ -3,17 +3,19 @@ use std::num::NonZeroU8;
...
+use secp256k1::rand::SeedableRng as _;
 use timeboost_crypto::{DecryptionScheme, TrustedKeyMaterial};
 use timeboost_utils::keyset::{KeysetConfig, NodeInfo, PrivateKeys, PublicDecInfo};
 use timeboost_utils::types::logging;
 
 #[derive(Clone, Debug, Parser)]
 struct Args {
     /// How many nodes should configuration contain?
-    #[clap(long, short)]
+    #[clap(long, short = 'n')]
```
twittner, line 18: Isn't `n` implied by `short`?

timeboost/src/binaries/mkconfig.rs
```diff
@@ -44,12 +50,41 @@ enum Mode {
 }
 
 impl Args {
-    fn mk_node_infos(&self, dec: &TrustedKeyMaterial) -> impl Iterator<Item = Result<NodeInfo>> {
-        iter::repeat_with(multisig::Keypair::generate)
-            .zip(iter::repeat_with(|| x25519::Keypair::generate().unwrap()))
+    fn mk_node_infos(
```
twittner, line 53: Why not always use `generate_with_rng`, e.g.

```rust
fn mk_node_infos(&self, dec: &TrustedKeyMaterial, seed: Option<u64>) -> impl Iterator<Item = Result<NodeInfo>> {
    let mut s_rng = StdRng::seed_from_u64(seed.map(|s| s.wrapping_pow(2)).unwrap_or_else(rand::random));
    let mut d_rng = StdRng::seed_from_u64(seed.map(|s| s.wrapping_pow(3)).unwrap_or_else(rand::random));
    iter::repeat_with(move || multisig::Keypair::generate_with_rng(&mut s_rng))
        .zip(iter::repeat_with(move || x25519::Keypair::generate_with_rng(&mut d_rng).unwrap()))
        .zip(&dec.2)
        .enumerate()
        .take(dec.2.len())
        .map(|(i, ((kp, xp), share))| {
            let i = i as u8;
            Ok(NodeInfo {
                sailfish_address: self.adjust_addr(i, &self.sailfish_base_addr)?,
                decrypt_address: self.adjust_addr(i, &self.decrypt_base_addr)?,
                producer_address: self.adjust_addr(i, &self.producer_base_addr)?,
                signing_key: kp.public_key(),
                dh_key: xp.public_key(),
                private: Some(PrivateKeys {
                    signing_key: kp.secret_key(),
                    dh_key: xp.secret_key(),
                    dec_share: share.clone(),
                }),
            })
        })
}
```

justfile
alxiong: True, do you have ideas how to properly generate these? 

cc @jparr721

alxiong: to i'll revert changes cloud configs, and we can update these in a separate PR.

timeboost/src/binaries/mkconfig.rs
alxiong: yes you are right, reverted in [722c69f](https://github.com/EspressoSystems/timeboost/pull/357/commits/722c69fed6e246cb460cb178158c81fbf3d29b4f)

timeboost/src/binaries/mkconfig.rs
```diff
@@ -44,12 +50,25 @@ enum Mode {
...
+            seed.map(|s| s.wrapping_pow(3)).unwrap_or_else(rand::random),
+        );
+        iter::repeat_with(move || multisig::Keypair::generate_with_rng(&mut s_rng))
+            .zip(iter::repeat_with(move || {
+                x25519::Keypair::generate_with_rng(&mut d_rng).unwrap()
+            }))
             .zip(&dec.2)
             .enumerate()
-            .take(dec.2.len())
+            .take(num_nodes as usize)
```
twittner, line 71: The `TrustedKeyMaterial` contains as many shares as there are nodes and since this method already takes the shares as a parameter, why not use its length?

PR #360, opened 2025-05-28
Multiple committees

twittner:
Pair round numbers with committee IDs to allow selection of the right committee when verifying signatures and certificates.

Also require vertex payloads to have a timestamp which will be used to derive the consensus timestamp when a round is delivered.

A `CommitteeVec` allows keeping multiple committees and indexing them by committee ID.


timeboost-sequencer/src/lib.rs
```diff
@@ -190,12 +191,17 @@ impl Sequencer {
 
             let rcf =
                 RbcConfig::new(cfg.sign_keypair.clone(), committee.clone()).recover(cfg.recover);
-            let rbc = Rbc::new(Overlay::new(net), rcf.with_metrics(rbc_metrics));
+
+            let rbc = Rbc::new(
+                5 * committee.size().get(),
```
jparr721, line 196: what is 5 here?

twittner: A magic number. The exact capacity should not matter.

timeboost-sequencer/src/decrypt.rs
```diff
@@ -196,8 +196,8 @@ struct Worker {
     label: PublicKey,
     /// overlay network to connect to other decrypters
     net: Overlay,
-    /// consensus committee
-    committee: Committee,
+    /// consensus committees
+    committees: CommitteeVec<2>,
```
lukeiannucci, line 200: can this also be the contant in coordinator.rs?

lukeiannucci (approved):
This looks good to me, i went through every file. just some questions that I have for my understanding

sailfish/src/coordinator.rs
```diff
@@ -79,25 +236,50 @@ where
...
+
+                if !(m.is_handover() || m.is_handover_cert()) {
+                    return Ok(Vec::new())
+                }
+
+                if self.previous_committees.contains(&c) {
+                    return Ok(Vec::new())
+                }
+
+                self.buffer.insert(m.signing_key().copied(), m);
```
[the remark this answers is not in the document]
twittner: We never accept connections from unknown peers. This can only originate from a peer we are already connected to in committee C1, who sends a handover message to us as a member of C2, when we do not know yet that we will be in C2.

PR #367, opened 2025-06-18
Expand committee handover test

twittner:
And add some small improvements.

### DIFF
--- a/sailfish-rbc/src/abraham/worker.rs
+++ b/sailfish-rbc/src/abraham/worker.rs
@@ -373,10 +373,7 @@ impl<T: Clone + Committable + Serialize + DeserializeOwned> Worker<T> {
         self.comm.remove(old.collect()).await?;
         self.comm.assign(Role::Active, committee.parties().copied().collect()).await?;
         self.config.committee_id = round.committee();
-        for (r, m) in &mut self.buffer {
-            if *r < round.num() {
-                continue
-            }
+        for (_, m) in self.buffer.range_mut(round.num() ..) {
             // Remove all messages from the old committee starting at round.
             m.map.retain(|d, _| d.round().committee() == round.committee())
         }


tests/src/tests/timeboost/handover.rs
```diff
@@ -126,14 +216,20 @@ async fn handover() {
                 select! {
                     cmd = cmd.recv() => match cmd {
                         Ok(Cmd::NextCommittee(t, a)) => {
-                            n.set_next_committee(t, a.committee().clone(), a).await.unwrap()
+                            n.set_next_committee(t, a.committee().clone(), a.clone()).await.unwrap();
+                            if a.committee().contains_key(&n.public_key()) {
+                                let d = repeat_with(Timestamp::now);
```
lukeiannucci, line 221: why is this the datasource?

sailfish-types/src/message.rs
```diff
@@ -328,14 +328,14 @@ impl<T: Committable> Message<T, Unchecked> {
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Payload<T: Committable> {
-    round: RoundNumber,
+    round: Round,
```
lukeiannucci, line 331: what was the reason for this change? is it just clean up? thats what it looks like to me but want to make sure im understanding

tests/src/tests/timeboost/handover.rs
```diff
@@ -180,20 +277,75 @@ async fn handover() {
...
         }
-        if c == committee2 {
-            committee_2_round = r
+        if r.committee() == c2 {
+            c2_round = r.num()
         }
-        if committee_2_round > 100.into() {
-            assert!(!committee_1_round.is_genesis());
+        if c2_round > 100.into() {
+            assert!(!c1_round.is_genesis());
```
lukeiannucci, line 293: is it also possible to check that the old members of the committee are still on committee id 0? or maybe this test is doing it and im missing it

lukeiannucci: also is it possible to somehow verify that `c1` isnt making progress but `c2` is?

lukeiannucci (approved):
lgtm! just some questions

sailfish-types/src/message.rs
twittner: It allows consumers to inspect the committee ID if desired. Currently used in the test, but I don't think it hurts to make it available.

PR #369, opened 2025-06-20
Add time variation in handover test.

twittner:
Delay command processing by some randomised delay and also delay the start of the next committee somewhat randomly.


lukeiannucci (approved):
lgtm

PR #372, opened 2025-06-23
Fix race condition during committee change.

twittner:
Consider the following sequence:

1. A `UseCommittee` action is created and the RBC layer activates a new committee in round `r`. It removes all messages of the old committee starting at `r`.
2. The node receives a proposal from another node which is in both committees for `r` + 1 but for the old committee.
3. Subsequently it receives a second proposal for `r` + 1 from the same node for the new committee.

At (3), a warning is issued that a node has made two proposal for the same round and the second proposal is rejected.

This commit adds the following constraint to `on_propose`: Proposals received for a committee older than the current one are rejected.

### DIFF
--- a/sailfish-consensus/src/lib.rs
+++ b/sailfish-consensus/src/lib.rs
@@ -13,7 +13,7 @@ use sailfish_types::math;
 use sailfish_types::{Action, Evidence, Message, NoVote, NoVoteMessage, Timeout, TimeoutMessage};
 use sailfish_types::{ConsensusTime, Handover, HandoverMessage};
 use sailfish_types::{DataSource, HasTime, Payload, Round, RoundNumber, Vertex};
-use tracing::{debug, error, info, trace, warn};
+use tracing::{Level, debug, enabled, error, info, trace, warn};
 
 pub use dag::Dag;
 pub use metrics::ConsensusMetrics;
@@ -246,13 +246,13 @@ where
     ///
     /// This means we did not receive a leader vertex in a round and
     /// results in a timeout message being broadcasted to all nodes.
-    pub fn timeout(&mut self, r: RoundNumber) -> Vec<Action<T>> {
+    pub fn timeout(&mut self, r: Round) -> Vec<Action<T>> {
         info!(node = %self.public_key(), round = %r, "internal timeout");
-        debug_assert_eq!(r, self.round());
-        let e = if r.is_genesis() {
+        debug_assert_eq!(r.num(), self.round());
+        let e = if r.num().is_genesis() {
             Evidence::Genesis
         } else {
-            self.evidence(r - 1)
+            self.evidence(r.num() - 1)
                 .expect("evidence for previous round exists")
         };
         let t = TimeoutMessage::new(self.committee.id(), e, &self.keypair);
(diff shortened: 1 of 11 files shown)


sailfish-consensus/src/lib.rs
```diff
@@ -723,7 +723,15 @@ where
         let r = v.round().data().num();
 
         if v.edges().any(|w| self.dag.vertex(r - 1, w).is_none()) {
-            debug!(node = %self.public_key(), vertex = %v, "not all edges are resolved in dag");
+            if enabled!(Level::DEBUG) {
+                let missing = v.edges().filter(|w| self.dag.vertex(r - 1, w).is_none());
+                debug!(
+                    node    = %self.public_key(),
+                    vertex  = %v,
+                    missing = ?missing.take(3).collect::<Vec<_>>(),
```
twittner, line 731: Unrelated quality of life improvement to help when studying log files.

sailfish-rbc/src/abraham/worker.rs
```diff
@@ -372,7 +383,7 @@ impl<T: Clone + Committable + Serialize + DeserializeOwned> Worker<T> {
             .copied();
         self.comm.remove(old.collect()).await?;
         self.comm.assign(Role::Active, committee.parties().copied().collect()).await?;
-        self.config.committee_id = round.committee();
+        self.round = (round, e);
```
twittner, line 386: The `Worker::config` field has been removed and the config is no longer modified by the worker. Instead the relevant fields are pulled on on construction and here we use the existing round information to also track the committee ID.

sailfish-rbc/src/abraham/worker.rs
```diff
@@ -622,10 +639,23 @@ impl<T: Clone + Committable + Serialize + DeserializeOwned> Worker<T> {
...
 
+        let our_committee_pos =
+            self.committees.position(self.round.0.committee())
+                .expect("current committee is member of committee vec");
+
+        let their_committee_pos =
+            self.committees.position(round.committee())
+                .expect("validated vertex committee is member of committee vec");
+
+        if their_committee_pos > our_committee_pos {
```
lukeiannucci, line 650: is it not enought to just check that `their_committee_pos` == 0?

sailfish-rbc/src/abraham/worker.rs
```diff
@@ -88,16 +94,18 @@ struct Messages<T: Committable> {
...
 impl<T: Committable> Messages<T> {
-    /// Get a message digest of this source, if any.
-    fn digest(&self, s: &PublicKey) -> Option<Digest> {
-        for (d, t) in &self.map {
-            if let Some(vertex) = &t.message.item {
-                if vertex.data().source() == s {
-                    return Some(*d);
+    /// Get message digests of this source.
+    fn digest(&self, s: &PublicKey) -> impl Iterator<Item = Digest> {
+        self.map
```
lukeiannucci, line 99: was there a bug previously as it would return the first item that matched the key?

lukeiannucci (approved):
looks good to me. just some questions

sailfish-rbc/src/abraham/worker.rs
twittner: No, the typical progression is as follows (a star '*' denotes the current committee):

1. $[C_1^*]$
2. $[C_2, C_1^*]$
3. $[C_2^*, C_1]$

PR #373, opened 2025-06-23
feat: add MRE encryption

alxiong:
Closes https://app.asana.com/1/1208976916964769/project/1209394409339229/task/1210580323499314

### This PR:

- Implement the simple Multi-Recipient Encryption (MRE) scheme in https://eprint.iacr.org/2025/1175.pdf
- added functional correctness test

### This PR does not:

- VESS on top will be implemented later in phase 2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,7 @@ ed25519-compact = "2.1.1"
 either = "1.13.0"
 ethereum_ssz = "0.9.0"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
+generic-array = { version = "^0.14.7", features = ["serde", "zeroize"] }
 multiaddr = "0.18.2"
 nohash-hasher = "0.2"
 parking_lot = "0.12.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ ed25519-compact = "2.1.1"
 either = "1.13.0"
 ethereum_ssz = "0.9.0"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
-generic-array = { version = "^0.14.7", features = ["serde", "zeroize"] }
+generic-array = { version = "0.14.7", features = ["serde", "zeroize"] }
 multiaddr = "0.18.2"
 nohash-hasher = "0.2"
 parking_lot = "0.12.3"
--- a/timeboost-crypto/src/lib.rs
+++ b/timeboost-crypto/src/lib.rs
@@ -1,4 +1,5 @@
 pub mod cp_proof;
+pub mod mre;
 pub mod sg_encryption;
 pub mod traits;
 
@@ -21,6 +22,7 @@ use std::{convert::TryFrom, num::NonZeroUsize};
 use traits::threshold_enc::ThresholdEncScheme;
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
+// TODO(alex): we should relocate these types to sg_encryption?
 #[derive(Copy, Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Nonce(u128);
 
--- a/timeboost-crypto/src/mre.rs
+++ b/timeboost-crypto/src/mre.rs
@@ -1,4 +1,4 @@
-//! Multi-recipient Encryption (MRE) allows encryption of a vector of message to all n parties.
+//! Multi-recipient Encryption (MRE) allows encryption of a vector of messages to all n parties.
 //! Proposed as MEGa in <https://eprint.iacr.org/2022/506>, this code implements the simplified
 //! variant in <https://eprint.iacr.org/2025/1175>.
 
@@ -24,8 +24,8 @@ pub struct MultiRecvCiphertext<C: CurveGroup, H: Digest> {
 
 impl<C: CurveGroup, H: Digest> MultiRecvCiphertext<C, H> {
     /// Extract the recipient-specific ciphertext
-    pub fn get_recipient_ct(&self, index: usize) -> Option<SingleRecvCiphertext<C, H>> {
-        self.cts.get(index).map(|ct| SingleRecvCiphertext {
+    pub fn get_recipient_ct(&self, index: usize) -> Option<Ciphertext<C, H>> {
+        self.cts.get(index).map(|ct| Ciphertext {
             epk: self.epk,
             ct: ct.clone(),
         })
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -70,6 +70,7 @@ ed25519-compact = "2.1.1"
 either = "1.13.0"
 ethereum_ssz = "0.9.0"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
+generic-array = { version = "^0.14.7", features = ["serde", "zeroize"] }
```
akonring, line 73: Is there a reason to use caret-requirement? I don't see this anywhere else in the code base. If we don't commit Cargo.lock could this not lead to non-deterministic builds (worst-case)?

timeboost-crypto/src/lib.rs
```diff
@@ -21,6 +22,7 @@ use std::{convert::TryFrom, num::NonZeroUsize};
 use traits::threshold_enc::ThresholdEncScheme;
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
+// TODO(alex): we should relocate these types to sg_encryption?
```
akonring, line 25: Yes. we should prolly do public re-export for better module org at some point.

timeboost-crypto/src/mre.rs
```diff
@@ -0,0 +1,220 @@
...
+            epk: self.epk,
+            ct: ct.clone(),
+        })
+    }
+}
+
+/// (Part of) [`MultiRecvCiphertext`] for a specific recipient.
+/// Only appropriate construction is [`MultiRecvCiphertext::get_recipient_ct()`]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SingleRecvCiphertext<C: CurveGroup, H: Digest> {
```
akonring, line 38: maybe just `Ciphertext`?

timeboost-crypto/src/mre.rs
```diff
@@ -0,0 +1,220 @@
...
+/// Multi-Recipient Encryption (MRE) in <https://eprint.iacr.org/2025/1175>
+///
+/// # Parameters
+/// - `recipients` and `messages` must be of the same length and implicitly indexed
+/// - `aad` is the associated data
+/// - `C` is the DL group, `H` is the choice of H_enc whose output space = message space
+///   - preprocess messages to pad them to proper length before passing in
+pub fn encrypt<'pk, 'msg, C, H, R>(
+    recipients: impl IntoIterator<Item = &'pk C::Affine>,
+    messages: impl IntoIterator<Item = &'msg Vec<u8>>,
```
akonring, line 52: curious about the advantages of explicit lifetimes compared to anonymous lifetimes here?

timeboost-crypto/src/mre.rs
```diff
@@ -0,0 +1,220 @@
...
+
+            // TODO(alex): use SIMD vectorized XOR when `std::simd` move out of nightly,
+            // or rayon as an intermediate improvement
+            let ct = Output::<H>::from_iter(k.iter().zip(msg).map(|(ki, m)| ki ^ m));
+            Ok(ct)
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+
+    // check if two iterators have left-overs (thus mismatched length)
+    match (pk_iter.next(), msg_iter.next()) {
```
akonring, line 100: nit: could consider collecting the inputs into Vecs upfront to allow early return on length mismatch and reuse them for parallelism with rayon. But don't think it matters for our n = 13.

timeboost-crypto/src/mre.rs
```diff
@@ -0,0 +1,220 @@
+//! Multi-recipient Encryption (MRE) allows encryption of a vector of message to all n parties.
```
akonring, line 1:
```suggestion
//! Multi-recipient Encryption (MRE) allows encryption of a vector of messages to all n parties.
```

akonring (approved):

timeboost-crypto/src/mre.rs
alxiong: I was hoping to disambiguate from `Ciphertext` from threshold decryption scheme.

timeboost-crypto/src/mre.rs
alxiong: i was trying to avoid collecting (which incur mem copy), thus wrote this slightly unconventional logic. 
if we are fine with collection, then I might as well just accept a slice `&[T]`. which would both have `.len()` etc. 

what do you prefer?

timeboost-crypto/src/mre.rs
alxiong:
```
 mre.rs    51  43 error           rustc [E0658]: anonymous lifetimes in `impl Trait` are unstable
 expected named lifetime parameter (eglot-check)
```

ah, don't think it will work

Cargo.toml
alxiong: done in [ef76aee](https://github.com/EspressoSystems/timeboost/pull/373/commits/ef76aee1f338d7449327adeb69f458533c1920c9)

timeboost-crypto/src/mre.rs
alxiong: done in [ef76aee](https://github.com/EspressoSystems/timeboost/pull/373/commits/ef76aee1f338d7449327adeb69f458533c1920c9)

timeboost-crypto/src/mre.rs
akonring: My mistake. Was merely trying to point out that we can do lifetime elision here (but maybe there is a reason you want to keep them explicit).
```suggestion
    recipients: impl IntoIterator<Item = &C::Affine>,
    messages: impl IntoIterator<Item = &Vec<u8>>,
```

timeboost-crypto/src/mre.rs
akonring: have a weak preference for the slice approach

timeboost-crypto/src/mre.rs
alxiong: your suggestion also won't work, you need to specify lifetime to pacify the compiler.

timeboost-crypto/src/mre.rs
alxiong: done in [a272bbc](https://github.com/EspressoSystems/timeboost/pull/373/commits/a272bbcdc1add99cd443a3a301ca897f12dde91c)

akonring: Yes, thanks.

PR #376, opened 2025-06-25
feat: Feldman VSS implementation

alxiong:
Blocked by #374 
Closes https://app.asana.com/1/1208976916964769/project/1209394409339229/task/1210580323499323?focus=true

### This PR:

- Extrapolate previous Lagrange interpolation logic into its own files, into two functions: one for normal `interpolate()` which is used in VSS reconstruction, the other is `interpolate_in_exponent()` which is used in threshold decryption. (and later in resharing)
- Implement Feldman VSS and added unit tests.

### DIFF
--- a/timeboost-crypto/src/feldman.rs
+++ b/timeboost-crypto/src/feldman.rs
@@ -34,17 +34,12 @@ impl<C: CurveGroup> VerifiableSecretSharing for FeldmanVSS<C> {
     type Secret = C::ScalarField;
     type SecretShare = C::ScalarField;
     type Commitment = Vec<C::Affine>;
-    type ShareProof = (); // unused
 
     fn share<R: Rng>(
         pp: &Self::PublicParam,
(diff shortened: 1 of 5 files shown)


timeboost-crypto/src/feldman.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    fn reconstruct(
+        pp: &Self::PublicParam,
+        shares: impl Iterator<Item = (usize, Self::SecretShare)>,
+    ) -> Result<Self::Secret, VssError> {
+        let shares = shares.collect::<Vec<_>>();
+        let n = pp.n.get() as usize;
+        let t = pp.t.get() as usize;
+        // input validation
+        if shares.len() != t {
```
akonring, line 106: Do we want to allow reconstruction with more than `t` shares? If not, then maybe `InsufficientShares` should more like `ShareCountMismatch`.

PR #377, opened 2025-06-26
Add espresso network client.

twittner:
Adds a client to Espresso network, submits certified blocks and verifies the namespace proof.


timeboost-builder/src/submit.rs
```diff
@@ -0,0 +1,94 @@
...
+        loop {
+            let Ok(header) = robusta::watch(&self.config.robusta, self.height, nsid).await else {
+                let d = delays.next().expect("delay iterator repeats");
+                sleep(d).await;
+                continue;
+            };
+            delays = delay_iter();
+            match self.client.verify(&header, cb).await {
+                Ok(()) => {
+                    self.height = Height::from(header.height() + 1);
```
lukeiannucci, line 68: i think its possible that the block heights might differ between submission and verification and im not sure of the best solution:
example:
node 0 submits for block 1 and is included at hotshot height 10
node 1 is waiting but the hotshot height is ahead say on 11

since node 1 already passed hotshot height 10, and will have to wait now and resend after the 30 second delay, node 0 was able to verify it quickly since they were the sender and their local hotshot height was behind node 1. but now node 0 is stuck waiting for node 1 to submit the next block then increases its height too far.

do you think this scenario makes sense?

Cargo.toml
```diff
@@ -98,12 +99,18 @@ tide-disco = "0.9.3"
...
 tokio-stream = "0.1.17"
+tokio-tungstenite = { version = "0.27.0", features = ["rustls-tls-webpki-roots", "url"] }
 tokio-util = "0.7.15"
 toml = "0.8.19"
 tonic = "0.13.1"
 tonic-build = { version = "0.13.1", features = ["prost"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter", "json"] }
 turmoil = "0.6.4"
+url = "2.5.4"
```
lukeiannucci, line 110: the docker file wont run, i think we need to add libcurl dependency after these lines in the timeboost.dockerfile:
```
# Non-root app container stage
FROM debian:bullseye-slim

WORKDIR /app
```

so adding:
```
RUN apt update && apt-get install -y libcurl4
```
might fix it

lukeiannucci (approved):
i think it looks good, the hotshot height issue might be a real problem. and maybe we should discuss a solution tomorrow, i noticed very slow submission times and i expect this is the issue

robusta/src/watcher.rs
```diff
@@ -0,0 +1,121 @@
...
+
+type Ws = WebSocketStream<MaybeTlsStream<TcpStream>>;
+
+pub async fn watch<H, N>(cfg: &Config, height: H, nsid: N) -> Result<Header, WatchError>
+where
+    H: Into<Height>,
+    N: Into<Option<NamespaceId>>,
+{
+    let height = height.into();
+    let nsid = nsid.into();
```
lukeiannucci, line 21: why is namespace id optional?

timeboost-builder/src/submit.rs
lukeiannucci: im actually unsure on this now, but there is definetly something strange going on

PR #378, opened 2025-06-26
Lookup handover evidence.

twittner:
In case the first round after handover times out, the only possible evidence is the handover certificate which is not currently considered when looking up evidence of the previous round.


lukeiannucci (approved):
nice find!

PR #385, opened 2025-07-01
feat: VESS without verifiability for phase 1

alxiong:
Part of https://app.asana.com/1/1208976916964769/project/1209394409339229/task/1210580323499317?focus=true

### This PR:

As a first step to build out DKG logic, let's first add this VESS logic and apis in.
Right now, we are not implementing the actual VESS scheme, but just a thin wrapper around FeldmanVSS and MultiReceiver encryption. Which is sufficient for phase 1. 

Reviewers should check that this API should be reasonably stable for phase 2 when we add verifiability. I use the `struct VessCiphertext` which currently only contains a MRE ciphertext, but later would be a sigma proof from the cut-and-choose proof.

### DIFF
--- a/timeboost-crypto/src/vess.rs
+++ b/timeboost-crypto/src/vess.rs
@@ -1,0 +1,217 @@
+//! Verifiable Encrypted Secret Sharing (VESS) schemes
+
+use std::num::NonZeroU32;
+
+use ark_ec::CurveGroup;
+use ark_serialize::{CanonicalDeserialize, serialize_to_vec};
+use ark_std::{
+    marker::PhantomData,
+    rand::{CryptoRng, Rng},
+};
+use sha2::Digest;
+use thiserror::Error;
+
+use crate::{
+    feldman::{FeldmanVss, FeldmanVssPublicParam},
+    mre::{self, MultiRecvCiphertext},
+    traits::dkg::{VerifiableSecretSharing, VssError},
+};
+
+/// Implementation of [Shoup25](https://eprint.iacr.org/2025/1175)
+///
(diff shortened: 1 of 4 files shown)


timeboost-crypto/src/vess.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+}
+
+/// Error related to VESS scheme
+#[derive(Error, Debug)]
+pub enum VessError {
+    #[error("mre failed: {0}")]
+    Mre(#[from] mre::MultiRecvEncError),
+    #[error("vss failed: {0}")]
+    Vss(#[from] VssError),
```
akonring, line 157: how does vss fail here?

akonring (approved):
LGTM 👍

timeboost-crypto/src/vess.rs
```diff
@@ -0,0 +1,217 @@
...
+    VSS: VerifiableSecretSharing<Secret = C::ScalarField>,
+{
+    /// repetition param, N in paper
+    num_repetition: usize,
+    /// verifier "open" subset S, M=|S| in paper
+    subset_size: usize,
+    vss_pp: VSS::PublicParam,
+    _group: PhantomData<C>,
+    _mre_hash: PhantomData<H>,
+    _vss: PhantomData<VSS>,
```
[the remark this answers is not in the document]
alxiong: you are right, removing it

timeboost-crypto/src/vess.rs
alxiong: removed in [5ba3799](https://github.com/EspressoSystems/timeboost/pull/385/commits/5ba37990f8f42e569343cdd046a0200fdff279df)

PR #387, opened 2025-07-01
Add support for block signer committee changes.

twittner:
Also ensure that `CommitteeVec` is always non-empty.

On top of #384.


timeboost/src/lib.rs
```diff
@@ -87,26 +87,28 @@ impl Timeboost {
...
                             } else {
                                 error!(node = %self.label, "failed to encode inclusion list")
                             }
                         } else {
-                            warn!(
-                                node  = %self.label,
-                                round = %o.round(),
-                                "no forwarder configured => dropping sequencer output"
-                            )
+                            warn!(node = %self.label, %round, "no forwarder => dropping output")
```
lukeiannucci, line 105: i might have missed a change, we dont produce a block now if no forwarder is set up?

timeboost-builder/src/produce.rs
```diff
@@ -395,18 +532,44 @@ impl Worker {
...
+            .iter()
+            .find_map(|(_, c)| (c.id() == i).then_some(c))
+    }
+
+    /// Get the committee corresponding to the given round number.
+    fn committee_of_round(&self, r: RoundNumber) -> Option<&Committee> {
+        self.committees
+            .iter()
+            .find_map(|(n, c)| (r >= *n).then_some(c))
+    }
```
akonring, line 572: Thought: At some point we should consolidate this into something like CommitteeVec (which is not currently round-aware). Seems like this logic is useful in other places (like decrypter).

timeboost-builder/src/produce.rs
```diff
@@ -26,23 +30,33 @@ type Result<T> = StdResult<T, ProducerError>;
...
     /// Block number counter.
     counter: BlockNumber,
 }
 
+/// Worker commands.
+enum Command {
+    /// Certify the given block.
+    Certify(BlockNumber, Block),
+    /// Prepare for the next committee.
+    NextCommittee(AddressableCommittee),
```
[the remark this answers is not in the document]
twittner: There is currently no code to trigger a committee change. Eventually the flow could be like this:

1. Next committee is detected by some component and reported to `Timeboost`.
2. `Timeboost` calls:
    - `Sequencer::set_next_committee`
    - `Producer::next_committee`
3. Once Sailfish makes the switch, the `Sequencer` would get an `Action::UseCommittee` result which it includes in its own output.
4. When `Timeboost` receives `Output::UseCommittee` from `Sequencer` it can call `Producer::use_committee`

timeboost/src/lib.rs
twittner: This PR only changed the wording but yes, without `Nitro` there is no block.

akonring (approved):
👍

PR #399, opened 2025-07-09
More block producer tweaks

twittner:
1. Move `NodeInfo` from `sailfish-consensus` to `sailfish-types` and use it in `Producer` to determine the lower bound for garbage collection.
2. Use `CommitteeVec` in `Producer` and always determine committee by ID.

### DIFF
--- a/sailfish-types/src/nodeinfo.rs
+++ b/sailfish-types/src/nodeinfo.rs
@@ -14,6 +14,7 @@ impl<T: Default + PartialOrd + Clone> NodeInfo<T> {
         }
     }
 
+    /// Store a value of a party.
     pub fn record(&mut self, k: &PublicKey, new: T) -> bool {
         let Some(i) = self.nodes.iter().position(|(p, _)| p == k) else {
             return false;
@@ -41,7 +42,7 @@ impl<T: Default + PartialOrd + Clone> NodeInfo<T> {
         true
     }
 
-    /// Gets the smallest round number the quorum of nodes has committed.
+    /// Gets the lower bound of the highest quorum interval.
     pub fn quorum(&self) -> T {
         debug_assert!(self.quorum <= self.nodes.len());
         self.nodes[self.quorum - 1].1.clone()
--- a/sailfish-types/src/nodeinfo.rs
+++ b/sailfish-types/src/nodeinfo.rs
@@ -1,25 +1,20 @@
 use multisig::{Committee, PublicKey};
-use sailfish_types::RoundNumber;
 
-/// Information about nodes.
 #[derive(Debug)]
-pub struct NodeInfo {
-    /// Associative list of public key to committed round number.
-    nodes: Vec<(PublicKey, RoundNumber)>,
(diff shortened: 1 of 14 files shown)


timeboost-builder/src/produce.rs
```diff
@@ -29,6 +29,9 @@ use crate::BlockProducerConfig;
 
 type Result<T> = StdResult<T, ProducerError>;
 
+const CAPACITY: usize = 128;
+const HISTORY: u64 = 100;
```
akonring, line 33: Any specific reason to have HISTORY=100? IIRC in Sailfish we only subtract committee.size() from max quorum height. Do we need extra buffer here?

sailfish-types/src/nodeinfo.rs
```diff
@@ -39,16 +34,16 @@ impl NodeInfo {
...
         debug_assert!({
-            let it = self.nodes.iter().map(|(_, r)| *r);
+            let it = self.nodes.iter().map(|(_, r)| r);
             it.clone().zip(it.skip(1)).all(|(a, b)| a >= b)
         });
 
         true
     }
 
     /// Gets the smallest round number the quorum of nodes has committed.
```
akonring, line 44:
```suggestion
    /// Gets the largest round number the quorum of nodes has committed.
```

timeboost-builder/src/produce.rs
```diff
@@ -228,21 +231,8 @@ struct Worker {
...
-    /// We keep one tracker per block info.
-    ///
-    /// The reason for that is that a `VoteAccumulator` corresponds to one
-    /// committee and the block info contains a committee ID that is used
-    /// to select the committee. The committee can therefore not be subject
-    /// to voting.
-    trackers: BTreeMap<BlockInfo, Tracker>,
-    /// We remember the voters and only allow one vote per party and block number.
-    voters: SmallVec<[KeyId; 16]>,
+    /// Quorum of block numbers to use with garbale collection.
```
akonring, line 234:
```suggestion
    /// Quorum of block numbers to use with garbage collection.
```

akonring (commented):
👍

sailfish-types/src/nodeinfo.rs
twittner: > the smallest round number a quorum of nodes has committed is 1 (in your example).

Yes, and the largest is $5$. :wink:

timeboost-builder/src/produce.rs
```diff
@@ -382,6 +352,7 @@ impl Worker {
         let msg = Message {
             info: Envelope::signed(info.clone(), &self.keypair),
             evidence: evi,
+            next: self.next_block.unwrap_or_default(),
```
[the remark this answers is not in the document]
twittner: Yes, `next` tracks the next block we want to deliver locally which is $\le$ to the block number in `info`. We may be able to send out several messages with increasing block numbers, but are still waiting for some smaller block number to have enough signatures. For GC we move the potentially smaller number.

PR #400, opened 2025-07-09
Dynamic committees for decryption phase

akonring:
### Context
To support dynamic committees, we need more than just the “handover” in consensus (Sailfish). Other components running on separate networks must also be made aware of committee changes so they can handle these transitions properly.

This is especially important during the decryption phase, where a committee switch should also trigger a Distributed Key Generation (DKG) event and Sailfish serves as the communication layer for this new key material.

Previous work on unifying the decryption and block building phases (Decrypter and Certifier) makes it possible to reuse the same committee switching logic across both. While further generalization of Decrypter and Certifier is desirable, some challenges remain:

1. The Decrypter uses evidence directly from Sailfish, while the Certifier must build this evidence independently. This is because multiple blocks can be produced per round, making Sailfish’s evidence model incompatible with block production.

2. Garbage collection in the Decrypter relies on a "leash" mechanism (back-pressure to Sailfish). In contrast, the Certifier lacks a direct connection between blocks and rounds, requiring the use of the lowest committed round quorum approach, similar to Sailfish.

Although these differences make complete unification challenging right now, aligning the committee change logic already improves readability and sets the stage for future convergence.

### Content
This PR finalizes support for dynamic committees in the decryption phase and prepares the groundwork for introducing DKG in the Decrypter. In particular, correctly integrating DKG is now feasible thanks to the addition of NextCommittee and UseCommittee events in the Decrypter.

Key changes:

1. Added a Round struct (comprising round_number and committee_id) to DecShareBatch. This includes the target committee_id so receivers can perform the correct threshold combination.

2. Implemented evidence checking against the correct local committee for each round.

3. Partitioned decryption shares for each round by committee_id. For hatching, we identify the first committee where threshold decryption is possible and discard shares from other committees for the same round.

### DIFF
--- a/timeboost-sequencer/src/decrypt.rs
+++ b/timeboost-sequencer/src/decrypt.rs
@@ -585,11 +585,10 @@ impl Worker {
         let Some(per_ct_opt_dec_shares) = self.dec_shares.get(&round) else {
             return Ok(None);
         };
-        let Some(committee) = self.round_committee(round) else {
+        let Some(committee) = self.round_committee(round).cloned() else {
             warn!(node = %self.label, %round, "no committee for hatching round");
             return Err(DecrypterError::NoCommitteeForRound(round));
         };
-        let c: Committee = committee.clone();
         if per_ct_opt_dec_shares.is_empty()
             || per_ct_opt_dec_shares.iter().any(|opt_dec_shares| {
                 let num_valid_shares = opt_dec_shares.iter().filter(|s| s.is_some()).count();
@@ -597,8 +596,8 @@ impl Worker {
 
                 // for valid decryption shares, as long as reaching f+1, we may try to hatch
                 // for invalid ones, we need 2f+1 to ensure consensus (to ignore invalid ciphertext)
-                num_valid_shares < c.one_honest_threshold().get()
-                    && num_invalid_shares < c.quorum_size().get()
+                num_valid_shares < committee.one_honest_threshold().get()
+                    && num_invalid_shares < committee.quorum_size().get()
             })
         {
             return Ok(None);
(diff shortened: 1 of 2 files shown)


timeboost-sequencer/src/decrypt.rs
```diff
@@ -505,17 +566,20 @@ impl Worker {
         let Some(per_ct_opt_dec_shares) = self.dec_shares.get(&round) else {
             return Ok(None);
         };
+        let Some(committee) = self.round_committee(round) else {
```
twittner, line 569:
```suggestion
        let Some(committee) = self.round_committee(round).cloned() else {
```
With this you do not need the explicit clone below.

timeboost-sequencer/src/lib.rs
```diff
@@ -305,6 +305,9 @@ impl Task {
                             let acts = self.sailfish.set_next_consensus(cons);
                             candidates = self.execute(acts).await?
                         }
+                        if let Err(err) = self.decrypter.next_committee(a).await {
+                            warn!(node = %self.label, %err, "decrypt next committee error");
```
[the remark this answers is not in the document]
akonring: fixed in [4cd4628](https://github.com/EspressoSystems/timeboost/pull/400/commits/4cd4628e69ed72c7867103f29c0e2c1bbfa81a0c)

timeboost-sequencer/src/decrypt.rs
akonring: thanks.
fixed in [abb68a0](https://github.com/EspressoSystems/timeboost/pull/400/commits/abb68a07ebdd6fbc1d2ffde09a7a51bc137fac1a)

PR #402, opened 2025-07-10
Init dealings as dkg bundles

akonring:
This PR adds DkgBundle as a special bundle type for carrying dealings.

### DIFF
--- a/timeboost-sequencer/src/decrypt.rs
+++ b/timeboost-sequencer/src/decrypt.rs
@@ -14,7 +14,7 @@ use timeboost_types::{DecryptionKey, InclusionList};
 use tokio::spawn;
 use tokio::sync::mpsc::{Receiver, Sender, channel};
 use tokio::task::JoinHandle;
-use tracing::{debug, error, trace, warn};
+use tracing::{debug, error, info, trace, warn};
 
 use crate::config::DecrypterConfig;
 
@@ -23,8 +23,10 @@ type DecShare = <DecryptionScheme as ThresholdEncScheme>::DecShare;
 type Ciphertext = <DecryptionScheme as ThresholdEncScheme>::Ciphertext;
 
 /// Command sent to Decrypter's background worker
-#[allow(dead_code)]
+#[allow(unused)]
 enum Command {
+    // request to inform the worker of DKG shares (dealings) in the inclusion list
+    Dkg(InclusionList),
     // request to decrypt all encrypted transactions inside the inclusion list
     Decrypt(InclusionList),
     /// Prepare for the next committee.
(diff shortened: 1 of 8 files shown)


twittner (approved):

timeboost-types/src/bundle.rs
```diff
@@ -295,6 +297,38 @@ impl Committable for SignedPriorityBundle {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct DkgBundle {
+    committee_id: CommitteeId,
+    transcript: Vec<u8>, // TODO: replace with VESS transcript (dealing)
```
alxiong, line 303: - you could directly use `struct VessCiphertext` that's already availabile in `vess.rs` 
- I'd add code doc on `DkgBundle` that this is a dealing from a DKG party, a dealing contains the Feldman commitment and encrypted ciphertext for _all_ other parties.

yapper/src/tx.rs
```diff
@@ -41,6 +41,10 @@ async fn send_bundle_to_node(
                 .send()
                 .await
         }
+        _ => {
+            warn!("Unsupported bundle variant");
```
alxiong, line 45: just checking, this is mostly for generating load for testing?

timeboost-types/src/candidate_list.rs
```diff
@@ -18,6 +20,7 @@ struct Inner {
     index: DelayedInboxIndex,
     priority: Vec<SignedPriorityBundle>,
     regular: Vec<Bundle>,
+    dkg: Option<DkgBundle>,
```
alxiong, line 23: why a single bundle instead of a vector,
isn't a CandidateList extrapolated from all the DAG vertex from a sailfish round? which may contains multiple dkg dealings?

timeboost-sequencer/src/decrypt.rs
```diff
@@ -129,10 +133,19 @@ impl Decrypter {
...
+        if incl.has_dkg_bundles() {
+            self.worker_tx
+                .send(Command::Dkg(incl.clone()))
+                .await
+                .map_err(|_| DecrypterDown(()))?;
+        }
+
         if incl.is_encrypted() {
             self.worker_tx
                 .send(Command::Decrypt(incl))
```
alxiong, line 151: also want to ask this: do we want the Decrypter to early return here? what if a `InclusionList` both `has_dkg_bundles()` and `is_encrypted`?

akonring: Yes, I think the details in the decrypter are still subject to change. I don’t think we need an early return here. The approach in this PR is that the Decrypter sends complete inclusion lists with both commands to the worker, and then it’s up to the worker to process the commands and extract what’s needed.

alxiong: >the Decrypter sends complete inclusion lists with both commands to the worker

got it, i think we could revisit this later. I'm also leaving TODO in my local changes on this point.

alxiong (approved):

PR #405, opened 2025-07-10
Use `grpc` connection in `Forwarder`

lukeiannucci:
Closes #<ISSUE_NUMBER>

### This PR:

Instead of managing inclusion list through our own tcp socket, we can utilize a grpc client/server pair for simplicity.

### DIFF
--- a/timeboost/src/forwarder/nitro_forwarder/worker.rs
+++ b/timeboost/src/forwarder/nitro_forwarder/worker.rs
@@ -3,14 +3,13 @@ use std::{iter::repeat, time::Duration};
 use multisig::PublicKey;
 use timeboost_proto::{forward::forward_api_client::ForwardApiClient, inclusion::InclusionList};
 use tokio::{sync::mpsc::Receiver, time::sleep};
-use tonic::{Status, transport::Channel};
+use tonic::transport::Channel;
 use tracing::warn;
 
 pub struct Worker {
     key: PublicKey,
     client: ForwardApiClient<Channel>,
     incls_rx: Receiver<InclusionList>,
-    pending: Option<InclusionList>,
 }
 
 impl Worker {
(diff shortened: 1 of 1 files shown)


timeboost/src/forwarder/nitro_forwarder/worker.rs
```diff
@@ -1,165 +1,54 @@
...
     }
 
     pub async fn go(mut self) {
         loop {
-            let connect = if let Some(f) = &mut self.connect_task {
-                Either::Right(f)
-            } else {
-                Either::Left(pending())
+            if let Some(incl) = self.pending.take() {
+                self.send(incl).await;
```
lukeiannucci, line 31: maybe we should somehow do exponential backoff on failures

timeboost/src/lib.rs
```diff
@@ -110,12 +110,22 @@ impl Timeboost {
...
+                                    .into_iter()
+                                    .map(|tx| proto::inclusion::Transaction {
+                                        encoded_txn: tx.encoded_2718(),
+                                        address: tx.address().as_slice().to_vec(),
+                                        timestamp: **tx.time(),
+                                    })
+                                    .collect(),
+                                    consensus_timestamp: timestamp.into(),
+                                    delayed_messages_read: 0,
+                            };
```
twittner, line 125: This should happen inside for `enqueue` I think.

timeboost/src/forwarder/nitro_forwarder.rs
```diff
@@ -23,17 +27,39 @@ impl Drop for NitroForwarder {
 
 impl NitroForwarder {
     pub async fn connect(key: PublicKey, addr: Address) -> Result<Self, Error> {
+        let uri = format!("http://{addr}");
+        let endpoint = Endpoint::from_shared(uri).map_err(|e| Error::InvalidUri(e.to_string()))?;
+        let chan = endpoint.concurrency_limit(1).connect().await?;
```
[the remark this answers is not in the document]
lukeiannucci: yes this is true. i can revert this

twittner (approved):

PR #415, opened 2025-07-16
fix: correct rand usage in frame-handling test

Fallengirl:
Replace non-existent rand::rng() with thread_rng() and random_range() with gen_range() to fix compilation error in the frame-handling test. This change maintains the same functionality while using the correct API from the rand crate


twittner:
Have a look at `rand` versions >= 0.9:

- `rand::rng`: https://docs.rs/rand/0.9.1/rand/fn.rng.html
- `rand::Rng::random_range`: https://docs.rs/rand/0.9.1/rand/trait.Rng.html#method.random_range

PR #419, opened 2025-07-20
Convert DKG output to Threshold DecryptionKey

alxiong:
Address the missing TODO in #417 

### This PR:

add the missing conversion after VESS is done at the end of the DKG procedure into the DecryptionKey required for the threshold scheme.

### DIFF
--- a/timeboost-types/src/decryption.rs
+++ b/timeboost-types/src/decryption.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 
 use anyhow::anyhow;
-use ark_ec::AffineRepr;
+use ark_ec::{AffineRepr, CurveGroup};
 use multisig::{Committee, KeyId};
 use rayon::prelude::*;
 use timeboost_crypto::{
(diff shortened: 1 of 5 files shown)


timeboost-types/src/decryption.rs
```diff
@@ -28,6 +33,38 @@ impl DecryptionKey {
...
+    /// - `committee_size`: size of the threshold committee
+    /// - `node_idx`: in 0..committee_size, currently same as KeyId
+    /// - `commitment`: the Feldman Commitment (see output of `ShoupVess::encrypted_shares()` and
+    ///   `FeldmanVss::share()`)
+    /// - `key_share`: the decrypted secret share from `ShoupVess::decrypt_share()`
+    pub fn from_dkg(
+        committee_size: usize,
+        node_idx: usize,
+        commitment: &<Vss as VerifiableSecretSharing>::Commitment,
+        key_share: <Vss as VerifiableSecretSharing>::SecretShare,
```
akonring, line 48: Wonder if these should be _vectors_ of commitments and keyshares. Otherwise, the client would need to aggregate them.

akonring (approved):
LGTM.

timeboost-types/src/decryption.rs
akonring: yes, that sounds good.

akonring: yes, LGTM. we could also implement the operation directly on the FeldmanCommitment type.

PR #420, opened 2025-07-22
sync crypto changes on DKG feature branch

alxiong:
Closes #<ISSUE_NUMBER>

Just trying to sync changes and improvements made to `timeboost-crypto` on `ax/dkg` to main, so that I can use it in other branches without deep conflict resolution later.

### This PR:

- Replicate all changes from ax/dkg branch but only for timeboost-crypto/ crate
- Add missing derive_more dependency to workspace
- Update prelude type names from DecryptionKey/EncryptionKey to DkgDecKey/DkgEncKey
- Fix import references in timeboost-utils and mkconfig binary

### DIFF
--- a/timeboost-crypto/src/lib.rs
+++ b/timeboost-crypto/src/lib.rs
@@ -134,7 +134,7 @@ pub struct KeyShare<C: CurveGroup> {
     index: u32,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Plaintext(Vec<u8>);
 
 #[serde_as]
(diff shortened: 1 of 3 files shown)


alxiong:
Another change I want to make: refactor `vess.verify_internal()`, @akonring has asked this question: https://github.com/EspressoSystems/timeboost/pull/392#discussion_r2211827996

So, the context is: there's lots of overlapping logic between `vess.verify()` and `vess.decrypt()`, thus I extract the overlaps into `.verify_internal()` which doesn't finish the whole Prover-Verifier interaction, but only finish until step 4.b in the paper -- therefore you can see an internal struct called `ProverStateUntilStep4b`. 

But this cause lots of error logs like since our `spongefish` transcript will complain if it's not fully read on drop. 

```
2025-07-24T02:29:20.749468Z ERROR spongefish::sho: Unfinished operations:
 [Absorb(2304)]    
```

So I want to modify the `verify_internal()` to cover all of interactive proof steps. the efficiency loss should be negligible to the decrypter.

alxiong:
Since #414 is already ready for review, no point merging this separately. I'm keeping this branch simply for my resharing work to be based on before 414 merged in.

PR #424, opened 2025-07-24
Add support for getting `Delayed Inbox Index`

lukeiannucci:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/timeboost-sequencer/Cargo.toml
+++ b/timeboost-sequencer/Cargo.toml
@@ -7,6 +7,10 @@ rust-version.workspace = true
 
 [dependencies]
 alloy-consensus = { workspace = true }
+alloy-primitives = { workspace = true }
+alloy-provider = "1.0.23"
+alloy-rpc-types = "1.0.23"
+alloy-sol-types = "1.3.0"
 arrayvec = { workspace = true }
 bimap = { workspace = true }
 bincode = { workspace = true }
@@ -18,6 +22,7 @@ ethereum_ssz = { workspace = true }
 metrics = { path = "../metrics" }
 multisig = { path = "../multisig" }
 parking_lot = { workspace = true }
+reqwest = { workspace = true }
 sailfish = { path = "../sailfish" }
 serde = { workspace = true }
 thiserror = { workspace = true }
--- a/timeboost-sequencer/src/delayed_messages.rs
+++ b/timeboost-sequencer/src/delayed_messages.rs
@@ -1,0 +1,81 @@
+use std::time::Duration;
+
+use alloy_primitives::{Address, U256};
+use alloy_provider::{Provider, RootProvider, network::Ethereum};
+use alloy_rpc_types::{BlockId, BlockNumberOrTag, Filter};
+use alloy_sol_types::{SolEvent, sol};
+use tokio::time::sleep;
+
+use crate::queue::BundleQueue;
+
+sol! {
+    event InboxMessageDelivered(uint256 indexed messageNum, bytes data);
+}
+
+pub struct DelayedMessages {
+    address: Address,
+    provider: RootProvider<Ethereum>,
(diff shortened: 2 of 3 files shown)


docker/yapper.Dockerfile
```diff
@@ -12,6 +12,8 @@ FROM debian:bullseye-slim
 
 WORKDIR /app
 
+RUN apt update && apt-get install -y libcurl4
```
jparr721, line 15: What's this for?

timeboost-sequencer/src/delayed_inbox.rs
```diff
@@ -0,0 +1,127 @@
...
+                .provider
+                .get_block(BlockId::Number(BlockNumberOrTag::Finalized))
+                .await
+            {
+                let finalized = b.header().number();
+                if finalized == last_finalized {
+                    sleep(INBOX_DELAY).await;
+                    continue;
+                }
+                if last_finalized == 0 {
```
jparr721, line 76: what is 0?

lukeiannucci: this is initial startup, i can make this constant with comment

docker/yapper.Dockerfile
lukeiannucci: yapper needed this dependency, i dont think its my pr that introduced it, but some previous pr broke the yapper build, so when i was testing i needed this to get it working again

timeboost-sequencer/src/delayed_inbox.rs
```diff
@@ -0,0 +1,127 @@
...
+                    last_finalized = finalized.saturating_sub(300);
```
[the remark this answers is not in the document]
lukeiannucci: ill make this a constant make we dont need this check, i just had it so we werent making some huge request from the beginning of the chain to the latest finalized in the chain

timeboost-sequencer/src/delayed_inbox.rs
lukeiannucci: i added better logic with constant and comment

akonring (approved):
LGTM 👍 

Wonder if it would be simpler and less error prone (and even possible) to instead interact with the local Nitro node and have it relay the call to Ethereum (using eth_call). 
In any case, this is likely outside the scope of this PR.

lukeiannucci:
> LGTM 👍
> 
> Wonder if it would be simpler and less error prone (and even possible) to instead interact with the local Nitro node and have it relay the call to Ethereum (using eth_call). In any case, this is likely outside the scope of this PR.

Yes, I think this is the idea, I am working on the Nitro side of things and since we have this in the config, I adjust everything and it seems to work fine with the nitro node. I can update the config once I have completed the nitro side of things.

justfile
```diff
@@ -81,20 +81,27 @@ run *ARGS:
...
-mkconfig_local NUM_NODES *ARGS:
+mkconfig_local NUM_NODES RPC_URL PARENT_CHAIN_ID PARENT_INBOX_ADDRESS *ARGS:
   cargo run --bin mkconfig -- -n {{NUM_NODES}} \
     --sailfish-base-addr "127.0.0.1:8000" \
     --decrypt-base-addr "127.0.0.1:10000" \
     --certifier-base-addr "127.0.0.1:11000" \
     --internal-base-addr "127.0.0.1:5000" \
+    --parent-rpc-url {{RPC_URL}} \
+    --parent-chain-id {{PARENT_CHAIN_ID}} \
+    --parent-ibox-contr-addr {{PARENT_INBOX_ADDRESS}} \
```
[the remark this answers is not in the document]
alxiong: see if you like https://github.com/EspressoSystems/timeboost/pull/414/commits/4513b977001f5a1302c36b56ec134f913cc82e3c

PR #431, opened 2025-07-31
DKG Catchup

akonring:
### Summary

This PR implements catchup behaviour for the decryption phase, enabling nodes that missed DkgBundles during protocol execution to recover and participate in threshold decryption.

###  Changes Made

This PR:
* Adds a Protocol for exchanging messages between Decrypter workers.
* Introduces DkgState, an enum representing the node’s decryption state, including the key if DKG has completed.
* Implements basic message handling and vote counting to collect the agreed-upon subset from remote nodes, allowing recovering nodes to reconstruct the decryption key.

### Motivation

Catch-up behavior is essential for nodes that did not receive sufficient DkgBundles via candidate lists, often because they invoked catch-up in Sailfish. In such cases, the node requests the required subset of DkgBundles from peers using the dedicated network for the decryption phase.

### Testing

* Re-enabled encrypted bundle support in transaction_order and block_order tests.
* Added a new test simulating a recovering node by sending DkgBundles to only 4 out of 5 nodes. The fifth node initiates catch-up and successfully reconstructs the decryption key before threshold decryption.

### Additional Notes

- Due to the resource-intensive nature of DKG, running timeboost in slower environments can lead to GC events clogging the queue between the Decrypter and its worker (see [this comment](https://github.com/EspressoSystems/timeboost/pull/414#issuecomment-3113584217)).
To address this, the Decrypter now emits GC events only after DKG completion, when an encryption key is available.

- Some sections of the logic are left as-is for now. They could benefit from refactoring or optimization. These changes are intentionally deferred to avoid premature optimization, especially with upcoming resharing support, which will likely require extensive modifications to DKG handling.

### DIFF
--- a/tests/src/tests/timeboost.rs
+++ b/tests/src/tests/timeboost.rs
@@ -128,8 +128,6 @@ where
 
 /// Generate random bundles at a fixed frequency.
 async fn gen_bundles(enc_key: ThresholdEncKeyCell, tx: broadcast::Sender<BundleVariant>) {
-    // allow time for DKG to settle so recovering nodes may catchup
-    sleep(Duration::from_secs(20)).await;
     loop {
         let Ok(b) = make_bundle(&enc_key) else {
             warn!("Failed to generate bundle");
--- a/tests/src/tests/timeboost.rs
+++ b/tests/src/tests/timeboost.rs
@@ -128,6 +128,8 @@ where
 
 /// Generate random bundles at a fixed frequency.
 async fn gen_bundles(enc_key: ThresholdEncKeyCell, tx: broadcast::Sender<BundleVariant>) {
+    // allow time for DKG to settle so recovering nodes may catchup
+    sleep(Duration::from_secs(20)).await;
     loop {
         let Ok(b) = make_bundle(&enc_key) else {
             warn!("Failed to generate bundle");
--- a/timeboost-sequencer/src/decrypt.rs
+++ b/timeboost-sequencer/src/decrypt.rs
@@ -591,7 +591,7 @@ impl Worker {
             .ok_or_else(|| DecrypterError::UnknownKey(src))?;
 
         let subset = match self.dkg_tracker.get(&committee_id) {
-            Some(acc) => acc.bundles(),
+            Some(acc) if acc.completed() => acc.bundles(),
             _ => {
                 return Err(DecrypterError::Dkg(format!(
                     "DKG incomplete for requested committee: {committee_id}",
--- a/timeboost-sequencer/src/decrypt.rs
+++ b/timeboost-sequencer/src/decrypt.rs
@@ -835,8 +835,11 @@ impl Worker {
     /// The node will always try to catchup with the help of remote nodes first.
     async fn dkg_catchup(&mut self) -> Result<()> {
         let req = Protocol::GetRequest(self.current);
+        // the round number is ignored by the recieving party, but we don't want to give an
+        // arbitrary value since `gc()` will probably clean it up too early. Thus, we put in
+        // an estimated round number using the `.oldest_cached_round()`.
         self.net
-            .broadcast(0, serialize(&req)?) // placeholder round value, ignored by receiver
+            .broadcast(self.oldest_cached_round().u64(), serialize(&req)?)
             .await
             .map_err(|e| DecrypterError::End(e.into()))?;
         Ok(())
--- a/timeboost-sequencer/src/decrypt.rs
+++ b/timeboost-sequencer/src/decrypt.rs
@@ -415,10 +415,6 @@ struct Worker {
     #[builder(default)]
     dkg_tracker: BTreeMap<CommitteeId, DkgAccumulator>,
 
-    /// Committees for which Dkg has already been completed.
-    #[builder(default)]
-    dkg_completed: BTreeSet<CommitteeId>,
-
     /// cache of decrypted shares (keyed by round), each entry value is a nested vector: an
     /// ordered list of per-ciphertext decryption shares. the order is derived from the
     /// ciphertext payload from the inclusion list `self.incls` of the same round
@@ -678,7 +674,6 @@ impl Worker {
 
                 self.enc_key.set(dec_sk.pubkey().clone());
                 self.dkg_state = DkgState::Completed(dec_sk);
-                self.dkg_completed.insert(committee.id());
                 info!(committee_id = %committee.id(), node = %self.label, "DKG finished (node successfully recovered)");
             }
         }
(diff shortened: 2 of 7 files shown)


timeboost-types/src/decryption.rs
```diff
@@ -180,27 +181,36 @@ impl DkgKeyStore {
 pub struct DkgAccumulator {
     store: DkgKeyStore,
     threshold: usize,
-    bundles: HashSet<DkgBundle>,
+    bundles: Vec<DkgBundle>,
+    complete: bool,
```
alxiong, line 185: may i ask why is this necessary? 

re `HashSet/Vec`: what's the benefit of using vector here?

re `complete`: i thought the `dkg_completion` status is already tracked outside this accumulator, in the decrypter?

timeboost-types/src/decryption.rs
```diff
@@ -210,35 +220,44 @@ impl DkgAccumulator {
...
 
 /// Represents a finalized subset of DKG bundles sufficient to combine.
-#[derive(Debug, Clone)]
-pub struct Subset<'a> {
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct Subset {
     committe_id: CommitteeId,
-    bundles: &'a HashSet<DkgBundle>,
+    bundles: Vec<DkgBundle>,
 }
```
alxiong, line 246: can we have `SubsetRef` and `Subset` where `SubsetRef` is our original definition which incurs less memory clone, and the latter is only used during DKG catchup, wdyt?

tests/src/tests/timeboost.rs
```diff
@@ -128,6 +128,8 @@ where
 
 /// Generate random bundles at a fixed frequency.
 async fn gen_bundles(enc_key: ThresholdEncKeyCell, tx: broadcast::Sender<BundleVariant>) {
+    // allow time for DKG to settle so recovering nodes may catchup
+    sleep(Duration::from_secs(20)).await;
```
alxiong, line 132: is this necessary? (feels hacky)

timeboost-sequencer/src/lib.rs
```diff
@@ -37,7 +37,7 @@ pub use config::{SequencerConfig, SequencerConfigBuilder};
 use crate::delayed_inbox::DelayedInbox;
 
 type Result<T> = std::result::Result<T, TimeboostError>;
-type Candidates = VecDeque<(RoundNumber, Evidence, Vec<CandidateList>)>;
+type Candidates = VecDeque<(RoundNumber, Evidence, BTreeMap<PublicKey, CandidateList>)>;
```
alxiong, line 40: also not sure why this is necessary

timeboost-sequencer/src/decrypt.rs
```diff
@@ -32,6 +34,19 @@ type DecShare = <DecryptionScheme as ThresholdEncScheme>::DecShare;
 type Ciphertext = <DecryptionScheme as ThresholdEncScheme>::Ciphertext;
 type DecSharesCache = BTreeMap<RoundNumber, HashMap<Round, Vec<Vec<Option<DecShare>>>>>;
 
+/// The message types exchanged during decryption phase.
+#[derive(Debug, Serialize, Deserialize)]
+enum Protocol {
+    /// A request to retrieve a subset identified by the committee id.
+    GetRequest(Round),
```
alxiong, line 41: Q: why not directly ask for `CommitteeId`? why is the round number relevant in DKG catchup?

timeboost-types/src/decryption.rs
akonring: 1. The vec preserves the ordering so that the subset can be exchanged/compared immediately.
2. Yes, but for resharing (out-of-scope for now) we might want to check for completion for individual accumulators.

tests/src/tests/timeboost.rs
akonring: Yes, this is a fine balance. If gen_bundles generates bundles too quickly then (if DKG is slow) it will have generated all bundles before the node catching up will see them. Thus, never triggers recovery.

timeboost-types/src/decryption.rs
akonring: Agreed — we should avoid unnecessary clones. That said, I’d also like to avoid premature optimization, especially with resharing just around the corner.

To be honest, I’m not entirely sold on the current semantics. Also, not sure the current use of subset warrants its own struct.

timeboost-types/src/decryption.rs
alxiong: re 1: make sense.

re 2: not sure what do you have in mind for "check individual accumulator". imo, there will be two status outside, `dkg_completion` and `reshare_completion`. 
I prefer removing this (as it's unused for now anyway, right?)

timeboost-types/src/decryption.rs
alxiong: fair enough, let's revisit this later.

timeboost-types/src/decryption.rs
akonring: Right, I see the redundancy you are referring to. Slightly prefer removing the `dkg_completed` set and just let the node consult its `dkg_tracker` (with the `completed` field) but don’t feel strongly about it - will just remove the field for now.

tests/src/tests/timeboost.rs
alxiong: sry I didn't follow why it will lead to "never trigger recovery".

alxiong:
I still find `GetRequest(round)` semantically strange. since Dkg Subset are not associated with any round number, but I assume you choose it because our `Overlay::broadcast(round, data)` requires a round number. 

but I find the code to be more readable and intuitive if we `GetRequest(CommitteeId)` which is what most people would expect to see. 
Changes made in [ee85494](https://github.com/EspressoSystems/timeboost/pull/431/commits/ee85494170d409aa3b58c35108b2a9e336fae656)

if you disagree, we can revert

timeboost-sequencer/src/decrypt.rs
akonring: Thinking about this again; believe we need the round to place the broadcast message in the "correct" bucket. If we use a placeholder like round=0, then we will not get the acknowledgement guarantees of the Overlay network because gc will remove the bucket immediately.

alxiong: interesting, i didn't know that. can you point me to the "bucket" behavior? meanwhile, feel free to revert [ee85494](https://github.com/EspressoSystems/timeboost/pull/431/commits/ee85494170d409aa3b58c35108b2a9e336fae656)

tests/src/tests/timeboost.rs
akonring: The `recovery` is only triggered when a node observes an inclusion list with encrypted bundles but does not posses a decryption key yet.
https://github.com/EspressoSystems/timeboost/pull/431/files#diff-ed0d08786758292ed50f0c87e2fb0f0534cc9a0975697c11c9a1afaec5977bdaR353-R355

timeboost-sequencer/src/decrypt.rs
```diff
@@ -683,10 +876,23 @@ impl Worker {
...
+                self.recover(incl.clone()).await?;
+                self.pending.insert(incl.round(), incl.clone());
+                self.dkg_state = DkgState::Recover(HashMap::default());
+                return Err(DecrypterError::DkgPending);
+            }
+            DkgState::Recover(_) => {
+                // we already initiated catchup; awaiting response from remote nodes.
+                self.pending.insert(incl.round(), incl.clone());
+                return Err(DecrypterError::DkgPending);
+            }
```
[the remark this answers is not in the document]
alxiong: it also sidestep this problem: https://github.com/EspressoSystems/timeboost/pull/431#discussion_r2250950274

PR #437, opened 2025-08-06
Integrate resharing logic with VESS

alxiong:
Closes #<ISSUE_NUMBER>

### This PR:

- Refactors VESS (Verifiable Encrypted Secret Sharing) to support key resharing: Extends the ShoupVess implementation to handle both initial DKG and key resharing scenarios through a unified interface
- Externalizes VSS public parameters: Removes vss_pp from ShoupVess struct parameters, now deriving them from Committee objects at runtime
- Adds comprehensive resharing APIs: Introduces encrypt_reshares(), verify_reshares(), and decrypt_reshare() methods alongside existing DKG methods
- Implements Mode enum: Distinguishes between DKG and Resharing operations internally with proper verification logic for each
- Adds resharing support to DecryptionKey: Implements from_resharing() method to construct keys from resharing dealings
- Updates all call sites: Migrates existing code to use the new externalized parameter design


timeboost-types/src/decryption.rs
```diff
@@ -63,6 +67,29 @@ impl DecryptionKey {
         Self::from_single_dkg(committee_size, node_idx, agg_key_share, &agg_comm)
     }
 
+    /// Construct all key material for the threshold decryption from Key resharing.
+    pub fn from_resharing<I>(
+        old_committee: &Committee,
```
akonring, line 72: This appears to be the only case where a client (a new committee member) requires the "description" of the previous committee. Given it is only used for input validation, I wonder if it is strictly necessary, as it introduces downstream dependencies.

alxiong: based on your Slack message, I assume info from old committee is necessary? and I can leave this as is?

akonring (approved):
👍 
Like the modularity (code reuse) introduced by the `Mode` enum.

timeboost-types/src/decryption.rs
```diff
@@ -63,6 +67,29 @@ impl DecryptionKey {
...
+        new_committee: &Committee,
+        recv_node_idx: usize,
+        dealings: I,
+    ) -> anyhow::Result<Self>
+    where
+        I: ExactSizeIterator<
```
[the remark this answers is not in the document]
alxiong: Well, it's better to start with more permissive trait bound, and then change to `ExactSizeIterator` (which is more restrictive) later if we need `.len()` later inside that logic. 

Consistency doesn't matter that much here.

PR #440, opened 2025-08-08
key management contract implmentation

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
- this pr adds a full foundry-based contract development workflow and rust integration for the key manager contract. 
- adds timeboost-contracts folder with solidity sources, foundry config, and contract tests.
- rust bindings are generated and exposed as a separate crate for use in the main code
-  ci is updated to build contracts, generate bindings, and run contract tests
- justfile now has commands for building contracts, generating bindings, and running contract tests

### This PR does not:
- this pr does not add contract interaction to the main node logic yet, nor does it implement advanced contract features beyond the key manager.

### Key places to review:
- contracts/src/keymanager.sol and test/keymanager.t.sol for contract logic and tests
- justfile for new contract commands
- ci workflow additions

### How to test this PR: 
contract unit tests: 
```sh
just test-contracts
```

- 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210330229095215


timeboost-contracts/rust/src/bindings/key_manager.rs
```diff
@@ -0,0 +1,6686 @@
+/**
```
lukeiannucci, line 1: is there a way that we can add this to `.gitignore` and somehow just have a build script to generate these bindings at compile time ie just running `cargo build`? like something similar to what we do in `/timeboost-proto/build.rs`? this way we dont need to add more `just` commands as well?

timeboost-contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    function createEncryptionKeyset(bytes memory _thresholdEncKey) external onlyOwnerOrManager returns (uint64 keysetId) {
+        uint64 thisKeysetId = nextKeysetId;
+        thresholdEncKey = _thresholdEncKey;
+        emit CreatedEncryptionKeyset(nextKeysetId, _thresholdEncKey);
+        nextKeysetId++;
+        return thisKeysetId;
+    }
+
+    function scheduleCommittee(uint64 _effectiveTimestamp, CommitteeMember[] calldata _members)
```
lukeiannucci, line 65: is the idea here even if only one committee member changes we will still pass in the old committee members?

timeboost-contracts/rust/src/bindings/key_manager.rs
lukeiannucci: yea for `timeboost-proto` we require users to have `protocompiler` installed with the change we made. lets get opinions from @alxiong @akonring @twittner before proceeding. i am fine either way, it was just a suggestion

timeboost-contracts/.github/workflows/test.yml
alxiong, line 1: didn't we already have contract CI in `build-and-test.yml`? 
we better centralized the CI declaration.

alxiong (commented):
I have a few comments in the notion design page after this PR, happy to discuss if you disagree. 

Maybe first update the contract then ping me again for review?


Meanwhile, I will try to do the following changes (@alysiahuggins)
- reorg: all contracts will be directly under `contracts/`, the `contracts/rust` will be moved up to the project root as `timeboost-contracts` as another member crate
- try the `build.rs` idea.

alxiong:
I'll try to fix the CI error in my next commit

lukeiannucci:
> I'll try to fix the CI error in my next commit

@alxiong, you might need to install foundry before the build step on all the different ci jobs now

README.md
```diff
@@ -45,6 +66,16 @@ Release build
...
 
+## Contracts Build
+```sh
+just build-contracts
+```
+
+## Contract Bindings
+```sh
+just gen-contract-bindings
+```
```
alxiong, line 77: btw, we can remove this now.

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+
+    struct Committee {
+        uint64 effectiveTimestamp;
+        uint64 id;
+        CommitteeMember[] committeeMembers;
+    }
+
+    event ScheduledCommittee(uint64 indexed id, uint64 effectiveTimestamp, uint64 membersCount, bytes32 membersHash, address indexed scheduledBy);
+    event SetThresholdEncryptionKey(bytes thresholdEncryptionKey, address indexed setBy);
+    event ChangedManager(address indexed oldManager, address indexed newManager, address indexed changedBy);
```
alxiong, line 24: first, names could be better, 
second, since we already store all those info in the contract storage, we don't need such informative event. 

```suggestion
    event NewCommittee(uint64 indexed id);
    event ThresholdEncKeyUpdated();
    event ManagerChanged(address indexed oldManager, address indexed newManager);
```

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * @notice This function is used to initialize the contract.
+     * @dev Reverts if the manager is the zero address.
+     * @dev Assumes that the manager is valid.
+     * @dev This must be called once when the contract is first deployed.
+     * @param initialManager The initial manager.
+     */
+    function initialize(address initialManager) external initializer {
+        if (initialManager == address(0)) {
+            revert InvalidAddress(initialManager);
```
alxiong, line 65: no need to return the invalid address imo. just `InvalidAddress()`, would you agree?

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+     * @return committeeId The id of the new committee.
+     */
+    function setNextCommittee(uint64 effectiveTimestamp, CommitteeMember[] calldata members)
+        external
+        virtual
+        onlyManager
+        returns (uint64 committeeId)
+    {
+        if (members.length == 0) {
+            revert InvalidCommitteeMembers();
```
alxiong, line 129: we can be more specific here and say `EmptyCommittee`

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+        if (committees.length > 0) {
+            uint64 lastTimestamp = committees[committees.length - 1].effectiveTimestamp;
+            if (effectiveTimestamp <= lastTimestamp) {
+                revert InvalidEffectiveTimestamp(effectiveTimestamp, lastTimestamp);
+            }
+        }
+
+        if (committees.length > type(uint64).max) revert CommitteeIdOverflow();
+
+        committeeId = uint64(committees.length);
```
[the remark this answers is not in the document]
alxiong: agree, we should simply have a separate monotonic counter

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+        uint256 hi = committees.length - 1;
+        while (lo < hi) {
+            uint256 mid = (lo + hi + 1) / 2;
+            if (committees[mid].effectiveTimestamp <= currentTimestamp) {
+                lo = mid;
+            } else {
+                hi = mid - 1;
+            }
+        }
+        return committees[lo].id;
```
alxiong, line 207: imo, we don't need to do binary search, just reverse search is good. because it's either the last, or second last.

we can also avoid comparing with the first, simply start with the last, and find backwards.

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+     * @dev Reverts if there is no next committee.
+     * @dev Assumes that committees are stored in ascending order of effective timestamp.
+     * @dev Assumes that committees are not deleted from the committee array.
+     * @return committeeId The next committee id.
+     */
+    function nextCommitteeId() public virtual view returns (uint64 committeeId) {
+        uint64 currCommitteeId = currentCommitteeId();
+        if (currCommitteeId == committees.length - 1) {
+            revert NoCommitteeScheduled(committees[currCommitteeId].effectiveTimestamp);
+        }
```
alxiong, line 221: we don't need to check this, since our committtee id is predictably calculated, just returns curId + 1, doesn't matter whether the next committee has been set or not.

contracts/src/KeyManager.sol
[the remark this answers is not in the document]
alxiong: sure, we can do that as well. 

I think we need a decision here: either we have minimal on-chain footprint but paired with some heavy event-syncing offchain querying services (like we did for espresso-network)
OR we store them directly in contract storage and very simple offchain connector code. 

if we deploy on ArbOne, maybe it's not that expensive, and much faster to develop an e2e demo.
But emitting as events definitely is more future proof as that's probably where we eventually end up after a few iterations.

cc @jbearer @alysiahuggins @twittner @jparr721 @lukeiannucci

contracts/README.md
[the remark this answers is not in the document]
alysiahuggins: also remove and agreed https://github.com/EspressoSystems/timeboost/pull/440/commits/4814d6cba1611be36d9992c7d864c3223e1bb624

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+    struct CommitteeMember {
+        bytes pubKey;
+        bytes secureChannelKey;
+        bytes dkgEncKey;
+        string networkAddress;
+    }
+
+    struct Committee {
+        uint64 effectiveTimestamp;
+        uint64 id;
```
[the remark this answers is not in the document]
alysiahuggins: updated here thanks https://github.com/EspressoSystems/timeboost/pull/440/commits/4814d6cba1611be36d9992c7d864c3223e1bb624

contracts/src/KeyManager.sol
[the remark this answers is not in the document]
alysiahuggins: during the standup we agreed to leave it as is with the members stored on chain @akonring @alxiong

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,224 @@
...
+
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import {UUPSUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+
+contract KeyManager is Initializable, OwnableUpgradeable, UUPSUpgradeable {
+    struct CommitteeMember {
+        bytes pubKey;
+        bytes secureChannelKey;
+        bytes dkgEncKey;
```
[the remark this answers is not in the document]
alxiong: `sigKey, dhKey, dkgKey` sounds good

alxiong (commented):
>Alysia: 
but would there be deletions between two committees? example committees 1,2, 3 - can committee 2 be deleted? 

we can assume that won’t happen, when we garbage collect ALL committee until a point (this is what timeboost’s `gc()` do in all cases, we don’t _selectively_ prune.

what I would do is:
- keep the current map you have now, `id -> Committee`
- store two private variable: `_nextCommitteeId` and `_oldestStoredCommitteeId` (for the map)
- restore the struct Committee (i.e. no extra prev, next id info)
- remove `headCommitteeId`
- rename `removeCommittee(id)` to `pruneUntil(id)`  or `removeUntil(id)` (In @jbearer's first draft, he design the per-id deletion api, but imo it's more sensible and consistent with timeboost's current gc behavior to prune until a point/epoch.)

(After communicating with @alysiahuggins, she agree to further simplify according to the plan above, just copying over here for visibility)

contracts/src/KeyManager.sol
```diff
@@ -0,0 +1,278 @@
...
+        returns (uint64 committeeId)
+    {
+        if (members.length == 0) {
+            revert EmptyCommitteeMembers();
+        }
+
+        // ensure the effective timestamp is greater than the last effective timestamp
+        if (nextCommitteeId > 0) {
+            uint64 lastTimestamp = committees[nextCommitteeId - 1].effectiveTimestamp;
+            if (effectiveTimestamp <= lastTimestamp || effectiveTimestamp > lastTimestamp + 10 minutes) {
```
alxiong, line 183:
```suggestion
            if (effectiveTimestamp <= lastTimestamp) {
```

no need to add the 10 min restriction. each epoch usually takes much longer than 10 min (i thought it was mostly a day or even longer.)

alxiong (approved):
contract looks right. Thanks!

PR #442, opened 2025-08-11
Some minor teaks and updates

twittner:
This PR is a hotchpotch of unrelated smaller changes:

1. Updates dependencies:
    - tonic
    - generic-array
2. Renames `ThresholdEncKeyCell`'s `wait` to `read` and fixes a race condition.
3. Only enqueues DKG bundles in the `Decrypter` is it has capacity.
4. Removes the unused gas estimator.

Regarding (3) an alternative approach would be the have a separate channel in `Decrypter` that is not subject to the same capacity constraints as inclusion lists.


timeboost-crypto/src/prelude.rs
```diff
@@ -105,18 +102,13 @@ impl ThresholdEncKeyCell {
...
-    /// Asynchronously waits for the key to become available, then returns it.
-    ///
-    /// If the key is already present, it is returned immediately.
-    /// Otherwise, the current task is suspended until `set()` is called.
-    ///
-    /// The returned key is a clone of the stored key.
-    pub async fn wait(&self) -> ThresholdEncKey {
+    pub async fn read(&self) -> ThresholdEncKey {
         loop {
+            let fut = self.notify.notified();
```
lukeiannucci, line 107: why this change?

alxiong: this eliminates a possible race condition where notification is sent after the `self.get()` and before the old `self.notify.notified().await`, in which case the await might be forever waiting. 

Good catch, Toralf!

timeboost-sequencer/src/lib.rs
```diff
@@ -288,12 +287,15 @@ impl Task {
...
-            self.next_dkg(&mut candidates, &mut pending_dkgs);
-            if !pending_dkgs.is_empty() {
-                tracing::debug!(num_bundles = %pending_dkgs.len(), "enqueuing dkg bundles");
-                if let Err(err) = self.decrypter.enqueue_dkg(&mut pending_dkgs).await {
-                    error!(node = %self.label, %err, "dkg enqueue error");
+            if pending.is_none() {
+                while let Some(dkg) = dkg_bundles.pop_front() {
+                    if !self.decrypter.has_capacity() {
+                        dkg_bundles.push_front(dkg);
+                        break;
```
alxiong, line 294: 1. why do we only process DKG bundles when `pending` is none? shouldn't we always try to make progress on processing DKG bundles?
2. what's the benefit of imposing the capacity on DKG processing? Say the decrypter don't have capacity for `worker_rx` receiving inclusion list anymore, why should this prevents the decrypter from sending Dkg command to the worker? 
(maybe checking the `self.decrypter.worker_tx.capacity()>0` is reasonable imo)

alxiong (commented):
Thanks @twittner for the improvement. I learned a lot from your better taste of code. 

only have one main question.

PR #443, opened 2025-08-11
Ensure header heights are consecutive.

twittner:
In a `Multiwatcher`, the individual watchers only make progress when a threshold has delivered a value. Each watcher is required to deliver consecutive and strictly monotonic block heights, otherwise the watcher is dropped and re-connects after some time, starting at the current lower bound of heights.


lukeiannucci (approved):
:+1:

PR #445, opened 2025-08-13
Implement HTTP API using axum.

twittner:
WIP


timeboost/src/api.rs
```diff
@@ -1,3 +1,128 @@
...
+#[derive(Debug, Clone, Builder)]
+pub struct ApiServer {
+    bundles: Sender<BundleVariant>,
+    enc_key: ThresholdEncKeyCell,
+    metrics: Arc<PrometheusMetrics>,
+}
+
+impl ApiServer {
+    pub fn router(self) -> Router {
+        let public = Router::new()
```
lukeiannucci, line 32: ok so the idea here is this server is handling the metrics requests and submission requests, so we dont need separte ports anymore, rather more specific routes? this makes a lot of sense, thanks for cleaning this up.

.github/workflows/build-and-test.yml
```diff
@@ -122,5 +122,3 @@ jobs:
       run: just run_demo -s /tmp/stamp --ignore-stamp --yapper -k test-configs/local-5.json
     - name: Run tests with late-start node
       run: just run_demo -l -s /tmp/stamp --ignore-stamp --yapper -k test-configs/local-5.json
-    - name: Run sailfish demo
```
lukeiannucci, line 125: i noticed you removed `until` on this script. Is my understanding correct we are removing it from the CI since we already have sailfish + timeboost demo, meaning it doesnt make sense to keep this around?

twittner: I have reverted that change. While the sailfish demo is a bit redundant, since the same code is tested via timeboost demo, it is nevertheless a good idea to test that the `run-sailfish-demo` script and `sailfish.rs` itself behave properly.

PR #455, opened 2025-08-19
Allow Decrypter to catch up beyond the initial epoch

akonring:
Closes #<ISSUE_NUMBER>


timeboost-sequencer/src/decrypt.rs
```diff
@@ -692,50 +689,42 @@ impl Worker {
...
         let guard = self.key_stores.read();
-        let Some(key_store) = guard.get(res.committee_id) else {
-            return Err(DecrypterError::NoCommittee(res.committee_id));
-        };
-        key_store
+        let current = guard
+            .get(res.committee_id)
+            .ok_or(DecrypterError::NoCommittee(res.committee_id))?;
+
+        let prev = (guard.len() == 2).then(|| guard.last().clone());
```
alxiong, line 705: is this correct? node will only have 2 keystore when it's running `Running` state and `NextCommittee` event happens and a new keystore is registered and resharing starts, right? In this case, `prev` should be `guard.first()` instead?

also when a node just started (i.e. in `HandoverPending` state), it doesn't really have `prev`, correct? (since it only store the current keystore locally)

alxiong (commented):
good work! We don't have any tests for late joining at later stages yet, right? Do we plan to add it in future PR or? 

also remember to update the comment in:
```rust
    /// A request for DKG subset has been received.
    async fn on_dkg_request_msg(
```
to indicate that it can respond catchup request for both dkg and resharing

timeboost-types/src/decryption.rs
```diff
@@ -347,182 +358,317 @@ impl KeyStore {
...
+            self.complete = true;
+
             Some(subset)
         } else {
             None
         }
     }
+
+    /// Create a new finalized accumulator directly from key store and subset.
+    pub fn from_subset(key_store: KeyStore, subset: DkgSubset) -> Self {
```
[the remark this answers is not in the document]
akonring: Think we need this method to add the subset to the accumulator (and Tracker) so we can use it immediately when we get a dkg_request_msg from another node.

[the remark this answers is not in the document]
alxiong: I see, you are right.

timeboost-sequencer/src/decrypt.rs
alxiong: > , last() should always return the oldest.

that's not true? [`last()`](https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html#method.last) gave the most recently added. 

> In my view, a newly started node (that wasn’t part of the initial epoch) will be initialized with the previous committee’s description,

i see, this is an implicit decision that I don't know about. 
thanks for clarifying

alxiong: I guess you can merge for now and if this logic is wrong we should detect it during your next PR when doing the integration test.

PR #457, opened 2025-08-20
Use externally provided block numbers.

twittner:
The current way to have separate block number counters in nodes can not provide consistent block numbers. If a counter misses just a single increment, the node is permanently out of step with the others. We could potentially do away with block numbers altogether and just deliver certified blocks in the order they are given to us, however this complicates evidence handling, catch-up and GC in certifiers.

The approach taken here is to require block numbers as input, i.e. as part of the block protobuf file. The block provider needs to ensure that block numbers are consistent, consecutive and strictly monotonically increasing.

One difficulty is that with external block numbers we do not have a genesis number where we can require no evidence of a previous block. The way this is resolved here is that every node is allowed to send one message without evidence, but all subsequent messages are required to contain evidence of the previous block certificate. This can only work if at least t+1 certifiers start with the same block number, otherwise they can not form a certificate. When starting initially 2t+1 timeboost nodes have to be up to create consensus on candidate lists and eventually producing an inclusion list and a block and their certifier components should receive those in order. If individual certifiers recover from a crash they will wait for evidence before sending their first message.

Depends on: https://github.com/EspressoSystems/timeboost-proto/pull/6

### DIFF
--- a/timeboost-builder/src/certifier.rs
+++ b/timeboost-builder/src/certifier.rs
@@ -1,7 +1,5 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::result::Result as StdResult;
-use std::sync::Arc;
-use std::sync::atomic::{AtomicU64, Ordering};
 
 use bon::Builder;
 use bytes::{BufMut, Bytes, BytesMut};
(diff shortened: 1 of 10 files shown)


timeboost-builder/src/certifier.rs
```diff
@@ -299,38 +317,58 @@ impl Worker {
...
-        };
+        }
 
-        self.send(block, info, evi).await
+        self.send(block, info, evidence).await
     }
 
     /// Broadcast the certification message to everyone.
-    async fn send(&mut self, block: Block, info: BlockInfo, evi: Evidence) -> Result<()> {
+    async fn send(&mut self, block: Block, info: BlockInfo, evi: Option<Evidence>) -> Result<()> {
```
lukeiannucci, line 366: why is this optional? as far as i can see, we always call `send` when we have evidence

PR #467, opened 2025-08-25
Use spawn_blocking.

twittner:
Some decrypt operations are very CPU heavy and block a thread for tens of milliseconds.

### DIFF
--- a/timeboost-crypto/src/vess.rs
+++ b/timeboost-crypto/src/vess.rs
@@ -23,6 +23,7 @@ use spongefish::{
 };
 use std::collections::{HashMap, VecDeque};
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


twittner:
> Generally, is <40ms operations considered too long as a blocking op?

There should be no real blocking code with `async`, to not interfere with other tasks on the same thread and tens of milliseconds is a lot.

> I definitely don't mind if switching to `spawn_blocking` has zero cost. but just wanna check if that's the case. iiuc, spawning new thread does add overhead to thread pool management.

There is a dedicated pool for those threads and they are kept for 10s if idle (the setting is configurable). I expect a constant and small number of those threads and also think they will be reused a lot, at least for the regular verify operations.

PR #469, opened 2025-08-26
update node config and sourcing public keys from contract

alxiong:
Step 2~4 of: https://www.notion.so/espressosys/Startup-Semantics-around-first-fetch-2572431b68e980d6b36bf4f4a090bbaa

### This PR:

The new config workflow is described in https://github.com/EspressoSystems/timeboost/tree/ax/step2/test-configs

The main changes are:
- update `mkconfig`: write to separate per-node configs, instead of a central keyfile
  - `KeysetConfig` is replaced by `NodeConfig` (per-node) and `CommitteeConfig` held by the manager corresponding to info in `committee.toml`
- introduce `register` binary, expected to be run by the key manager to write to contract the next committee info
- update `timeboost`, `sailfish` binary on how they construct `peer_hosts_and_keys`: read from contract instead of loading from central keyfile
- update all affected just commands and docker build
  - changes made to `scripts/*` is worth reviewing, we spawn off an `anvil` test blockchain, run `deploy` and `register` on it, and keep it alive while we run timeboost with the deployed contract address to read peer info from.

Some side effect changes:

> > - unify how we "blackbox serialize" crypto struct by introducing `timeboost_utils::Blackbox` which will encode (bs58 on bincode bytes) data into string, which hides all internal structure of the struct. more rationale in [7d913c4](https://github.com/EspressoSystems/timeboost/pull/469/commits/7d913c43b82ea6d55f01dbfdc43924fd607c7bec)
> > - remove all the boilerplate of `to_bytes()` and `try_from_str()` on `timeboost-crypto`'s struct (previously we use them to do exactly what Blackbox abstract away through its `.encode()` and `.decode()` api), so now crypto code is much cleaner and don't have two ways to serialize to bytes (one is serde's Serialize, one is to_bytes(), now the latter is removed)
>
> 👆 this is outdated, Toralf correctly point out in https://github.com/EspressoSystems/timeboost/pull/469#discussion_r2307476761 that `Blackbox` undesirably erase type info, and he wrote much cleaner and idiomatic code via `Bs58Bincode<T>(T)`. 
changes in [6a4edd4](https://github.com/EspressoSystems/timeboost/pull/469/commits/6a4edd4c64fcd7321f7567149f03c4414c29f1fd)

☝️ this is further dropped, Toralf proposed a clean solution: [1db7de0](https://github.com/EspressoSystems/timeboost/pull/469/commits/1db7de078e219afbd579c93863e69663ec7b0c9f), simply via `serde(transparent)` and `to_bytes()` on the struct that we need to hide internals. 

cc @akonring


test-configs/c0/committee.toml
```diff
@@ -0,0 +1,37 @@
...
+sailfish_address = "127.0.0.1:8003"
+
+[[members]]
+signing_key = "i5ccK9sZFtdWpB6rp4deJKvb8UNMCYhxj7G2sGBAsdam"
+dh_key = "6czLW5NW4C74eQUBdzwHGmUzA5B8vbWpuH6amZHXtPnq"
+dkg_enc_key = "8t7p3VfTCMXxEAPrcLzMHiANGmEd2Y5aLkKAgrpe5ZyEmgGm2MR1ypweLbCw5RdXySe"
+sailfish_address = "127.0.0.1:8004"
+
+[[members]]
+signing_key = "h7qA6fzJMhYLotwxNRGGBEDqUkaNpCJ8cBeCGQuAumTq"
```
lukeiannucci, line 34: i see there are 6 committee members in here but only 6 node tomls.

.github/workflows/build-and-test.yml
```diff
@@ -195,7 +195,7 @@ jobs:
           --init &
         cd ..
     - name: Run test timeboost with nitro sequencer
-      run: just run_demo -s /tmp/stamp --ignore-stamp -k test-configs/local-2.json --rounds 10000 --yapper --nitro
+      run: just run_demo -s /tmp/stamp --ignore-stamp -c test-configs/c0 --rounds 10000 --yapper --nitro
```
lukeiannucci, line 198: i will have to rethink this, i was only spinning up two nodes because we also spin up 2 nitro sequencers and a parent geth node in the step before im not sure the ci will be able to handle 5 nodes and 5 sequencers

lukeiannucci: let me know what you think about this:
https://github.com/EspressoSystems/timeboost/pull/471

i had to add another committee to be registered (also nodes need the proper inbox contract address and also added nitro address). You can pretty much refactor it however you want (i really dislike bash). But youll notice since the spinning up this `test-node` in ci this will spin up a geth node for us so we dont need anvil to be started, we can deploy directly to this geth node.

test-configs/README.md
```diff
@@ -1,9 +1,21 @@
...
+
+To generate configs for all nodes in a new committee:
 
 ``` sh
-just mkconfig_local 5 --seed 42 > test-configs/local-5.json
-just mkconfig_local 13 --seed 42 > test-configs/local-13.json
-just mkconfig_docker 5 --seed 42 > test-configs/docker.json
+# see mkconfig.rs Args or `mkconfig --help` for more options
+just mkconfig 5
+just mkconfig 13 --nitro-addr "0xabc"
```
lukeiannucci, line 17: im a bit confused on what `nitro-addr` is here. isnt this going to be some network address? more than likely always something on the localhost?

docker-compose.yml
```diff
@@ -150,7 +151,7 @@ services:
         "--tps",
         "1",
         "--keyset-file",
-        "docker.json",
+        "committee.json",
```
lukeiannucci, line 154: wont this be `committe.toml` now? also our CI doesnt run this docker compose file as of now so i suspect there may be more tedious work to be done. i think if it doesnt run its fine (as long as docker builds) then i can help get this to work in a subsequent pr.

justfile
```diff
@@ -84,33 +84,33 @@ run *ARGS:
...
-    --parent-ibox-contr-addr {{PARENT_INBOX_ADDRESS}} \
-    --mode "increment-port" {{ARGS}} | jq
-
-mkconfig_docker NUM_NODES *ARGS:
-  just mkconfig_docker_full {{NUM_NODES}} "https://theserversroom.com/ethereum/54cmzzhcj1o/" 1 "0x4dbd4fc535ac27206064b68ffcf827b0a60bab3f" {{ARGS}}
+mkconfig NUM_NODES *ARGS:
+  #!/bin/bash
+  for i in $(seq 0 $(({{NUM_NODES}} - 1))); do \
+    echo "mkconfig for node $i"; \
+    just mkconfig_full "http://127.0.0.1:8545" 31337 "0x4dbd4fc535ac27206064b68ffcf827b0a60bab3f" \
```
lukeiannucci, line 91: just noting here for the default anvil node this address wont be correct, hence the delayed inbox index will never get updated. i dont think this will cause any issues with our ci or local runs

timeboost-sequencer/src/delayed_inbox.rs
```diff
@@ -24,26 +25,24 @@ sol! {
...
         node: PublicKey,
         cfg: &ChainConfig,
         queue: BundleQueue,
     ) -> Result<Self, Error> {
         let url = cfg.parent_chain_rpc_url();
         let parent_chain_id = cfg.parent_chain_id();
-        let provider = RootProvider::<N>::connect(url)
-            .await
-            .map_err(|e| Error::RpcError(e.to_string()))?;
+        let provider = ProviderBuilder::new().connect_http(url.clone());
```
lukeiannucci, line 45: just curious what happens here now if it cant connect to the parent chain?

timeboost-utils/src/keyset.rs
```diff
@@ -5,61 +5,93 @@ use cliquenet::Address;
...
-pub struct KeysetConfig {
-    pub keyset: Vec<NodeInfo>,
+/// Config for each node, containing private keys, public keys, chain_config, network addresses etc
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct NodeConfig {
+    pub net: NodeNetConfig,
+    pub keys: NodeKeyConfig,
+    pub chain_config: ChainConfig,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub nitro_addr: Option<Address>,
```
lukeiannucci, line 19: i wonder if this should be apart of `NodeNetConfig` maybe @twittner has thoughts?

lukeiannucci (commented):
i think the refactoring looks great so far! thanks for the great work on this!

test-configs/README.md
alxiong: I'm not sure what that config is either. just leave available as I found it

timeboost-sequencer/src/delayed_inbox.rs
lukeiannucci: ah cool, the next like below will handle it. thanks for for the explanation!

test-configs/README.md
lukeiannucci: i guess im asking is if the readme should we pass in a sample address with the flag to make it more clear?

.github/workflows/build-and-test.yml
alxiong: that looks great!

test-configs/README.md
alxiong: ah, you are right. This is a network address, I will fix it.

test-configs/c0/node_0.toml
```diff
@@ -0,0 +1,22 @@
+[net]
+sailfish_address = "127.0.0.1:8000"
+decrypt_address = "127.0.0.1:9000"
+certifier_address = "127.0.0.1:10000"
+internal_address = "127.0.0.1:11000"
+
+[keys]
+signing_key = "eiwaGN1NNaQdbnR9FsjKzUeLghQZsTLPjiL4RcQgfLoX"
+dh_key = "AZrLbV37HAGhBWh49JHzup6Wfpu2AAGWGJJnxCDJibiY"
+dkg_enc_key = "8t9PdQ61NwF9n7RU1du43C95ndSs6jn2EM7gRCfutVo2YXh6dyXAJiEWhpfYtPUv9gK"
```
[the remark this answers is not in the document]
alxiong: i agree, to achieve this, I want to refactor the serde approach timeboost-crypto is taking. Give me some time

justfile
alxiong: fixed in [915bb7c](https://github.com/EspressoSystems/timeboost/pull/469/commits/915bb7c3146eb6ddf009c3cfe4d754a26b67ae13), with the new just command, we won't make mistake accidentally i think.

timeboost-contract/src/binaries/register.rs
```diff
@@ -0,0 +1,86 @@
...
+    #[clap(short, long)]
+    url: Url,
+
+    /// The contract address of deployed KeyManager.sol proxy
+    #[clap(short, long)]
+    key_manager_addr: Address,
+
+    /// Path to the committee.toml config for the next committee
+    #[clap(short, long)]
+    config_file: PathBuf,
```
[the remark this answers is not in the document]
alxiong: done in [a2f9381](https://github.com/EspressoSystems/timeboost/pull/469/commits/a2f9381367b5f351ce24a0ecc9bce7f8f362dd1e)

test-configs/c0/node_0.toml
```diff
@@ -0,0 +1,22 @@
+[net]
+sailfish_address = "127.0.0.1:8000"
+decrypt_address = "127.0.0.1:9000"
+certifier_address = "127.0.0.1:10000"
+internal_address = "127.0.0.1:11000"
```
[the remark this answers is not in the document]
alxiong: done in [5efe3c0](https://github.com/EspressoSystems/timeboost/pull/469/commits/5efe3c0ec6e54ba5bb4e81689ac62697334e92ed)

timeboost-utils/src/lib.rs
```diff
@@ -61,3 +62,67 @@ pub fn select_peer_hosts(
         Box::new(keyset.iter().take(keyset.len())) as Box<dyn Iterator<Item = _>>
     }
 }
+
+/// Sometimes we don't want to reveal the struct inner fields when serializing.
+/// This intermediate type holds the bs58::encode(bincode::encode(T)) string that treats
+/// the type `T` as a `Serialize + Deserialize` blackbox.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Blackbox(String);
```
twittner, line 70: I don't fully understand the motivation here. Is `Blackbox` meant to be a type that (de-)serializes from/to bs58+bincode? Why does it hold a `String`?

test-configs/c0/node_0.toml
```diff
@@ -0,0 +1,22 @@
...
+[keys.dkg]
+secret = "AgrGYiNQMqPpLgwPTuCV5aww6kpcoAQnf4xuFukTEtkL1"
+public = "8t9PdQ61NwF9n7RU1du43C95ndSs6jn2EM7gRCfutVo2YXh6dyXAJiEWhpfYtPUv9gK"
+
+[chain_config]
+parent_chain_id = 31337
+parent_chain_rpc_url = "http://127.0.0.1:8545/"
+parent_ibox_contr_addr = "0x4dbd4fc535ac27206064b68ffcf827b0a60bab3f"
+parent_block_tag = "finalized"
+key_manager_contr_addr = "0x2bbf15bc655c4cc157b769cfcb1ea9924b9e1a35"
```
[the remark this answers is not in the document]
alxiong: yes, this is also deployed on parent_chain

timeboost-utils/src/lib.rs
alxiong: So if we serialize a struct in crypto, say EncryptionKey used for dkg, you will get:

```toml
[keys.dkg.pubilc]
u = "..."

[keys.dkg.secret]
alpha = "..."
```

revealing its internals. So we choose to serialize these T as `bs58::encode(bincode::encode(T)): String`, 
basically I use `Blackbox` to replace https://github.com/EspressoSystems/timeboost/blob/main/timeboost-utils/src/keyset.rs#L67-L95

scripts/run-timeboost-demo
```diff
@@ -111,21 +119,27 @@ if $nitro_enabled; then
...
+echo "=== Contract deployment completed ==="
+
 pids=()
-for (( i=0; i<$nodes; i++ )); do
+i=0
+for f in "$config_dir"/node_*.toml; do
     cmd=(target/release/timeboost
-        --id $i
+        --committee-id 0
+        --config-file "$f"
```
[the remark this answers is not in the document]
alxiong: yah. i also realize that. thanks! fixed in [0387291](https://github.com/EspressoSystems/timeboost/pull/469/commits/03872918e365e72340ceade00a8985b2421b3f89)

test-configs/c0/node_0.toml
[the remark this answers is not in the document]
alxiong: done in [8501a26](https://github.com/EspressoSystems/timeboost/pull/469/commits/8501a26a7b6fe6091c70266e692564e8c4cd8ea2)

test-configs/c0/node_0.toml
```diff
@@ -0,0 +1,22 @@
+[net]
+sailfish = "127.0.0.1:8000"
```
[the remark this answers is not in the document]
alxiong: renamed in [8501a26](https://github.com/EspressoSystems/timeboost/pull/469/commits/8501a26a7b6fe6091c70266e692564e8c4cd8ea2)

timeboost-utils/src/lib.rs
alxiong: I added more rationale in [7d913c4](https://github.com/EspressoSystems/timeboost/pull/469/commits/7d913c43b82ea6d55f01dbfdc43924fd607c7bec), see if you find it reasonable. 

https://github.com/EspressoSystems/timeboost/blob/7d913c43b82ea6d55f01dbfdc43924fd607c7bec/timeboost-utils/src/lib.rs#L70-L93

justfile
```diff
@@ -84,33 +81,39 @@ run *ARGS:
...
     --parent-rpc-url {{RPC_URL}} \
     --parent-chain-id {{PARENT_CHAIN_ID}} \
     --parent-ibox-contr-addr {{PARENT_INBOX_ADDRESS}} \
-    --mode "increment-port" {{ARGS}} | jq
-
-mkconfig_docker NUM_NODES *ARGS:
-  just mkconfig_docker_full {{NUM_NODES}} "https://theserversroom.com/ethereum/54cmzzhcj1o/" 1 "0x4dbd4fc535ac27206064b68ffcf827b0a60bab3f" {{ARGS}}
+    {{ARGS}}
+
+mkconfig_docker:
```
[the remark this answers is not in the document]
alxiong: wdyt of: [d466258](https://github.com/EspressoSystems/timeboost/pull/469/commits/d466258010fc4f27e69090d977f486a46e674b66)

timeboost-utils/src/lib.rs
twittner: Why do `DkgEncKey` and `DkgDecKey` not implement `Serialize`/`Deserialize` directly in a way that is compact and readable, e.g. for human consumption with base 58?

The problem with `Blackbox` is that it erases all type information. Having this in a larger struct means one has to cast at runtime to some specific type. If some specific encoding should be used and serde's `with` attribute is not desired, why not make it a generic types, e.g. `Bs58Bincode<T>(T)`?

lukeiannucci (commented):
this looks good to me now! thanks alex! will hold off on approving for now until others also take another pass

akonring (commented):
Did a quick pass—configs look neat 👍.
Looking ahead, we might eventually want to extend the contract interaction (syncing) logic so a node may fetch details for both the current and next committee, enabling handovers. That said, this feels out of scope for this already substantial contribution.

test-configs/c0/committee.toml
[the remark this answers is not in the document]
alxiong: I can take no credit from it. The taste of how it should look like is mostly instructed by @twittner. My original toml looks pretty ugly.

timeboost-utils/src/lib.rs
alxiong: adopted your code in [6a4edd4](https://github.com/EspressoSystems/timeboost/pull/469/commits/6a4edd4c64fcd7321f7567149f03c4414c29f1fd)

timeboost/src/binaries/mkconfig.rs
```diff
@@ -76,71 +89,98 @@ enum Mode {
...
-                dh_key: xp.public_key(),
-                enc_key: timeboost_crypto::prelude::DkgEncKey::from(&dkg_sk),
-                private: Some(PrivateKeys {
-                    signing_key: kp.secret_key(),
-                    dh_key: xp.secret_key(),
-                    dec_key: dkg_sk,
-                }),
-                nitro_addr: self.nitro_addr.clone(),
+        if !self.output.is_dir() {
+            bail!("--output only accepts valid directory path");
```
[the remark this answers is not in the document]
alxiong: fixed as part of [2148f92](https://github.com/EspressoSystems/timeboost/pull/469/commits/2148f92cb6a55a1344a69157b5a1c8ef1c0bb7ad)

alxiong:
re: @akonring 
> extend the contract interaction (syncing) logic so a node may fetch details for both the current and next committee, enabling handovers.

yes, imo, we should have another system paramter that specifies "how many rounds/seconds prior to `NextCommittee` should a node fetch from the contract about the committee info" and retry until succeed and then enter `NextCommittee` proposal.

alxiong:
@lukeiannucci do you know why the restriction on `nitro.address` there? why must they be consecutive?

lukeiannucci:
> @lukeiannucci do you know why the restriction on `nitro.address` there? why must they be consecutive?
> 
> (cuz I may need to edit my `mkconfig` to fix that, currently the magic port expansion logic for `nitro` is following public and decrypter etc. so need to know the source of the constraint here)

@alxiong 
there is not reason in particular, when i forked the `test-node` repository i decided to make them connsecutive (see:
https://github.com/EspressoSystems/decentralized-timeboost-nitro-testnode/blob/231efad2f890e692ec7716a48f69848ac528c6ca/docker-compose.yaml#L156-L200) i can also adjust this to match your mkconfig, so i dont think its a big deal, or we can set up the repository to pass in the ports based on the mkconfig in a future pr here:
https://github.com/EspressoSystems/timeboost/blob/1deb25a8d081171e59d302c0757e21c46a28c85e/.github/workflows/build-and-test.yml#L187-L196

for now i think your mkconfig is fine and dont need to modify it.

alxiong:
got it, would be nice to offer `--port` options on `test-node.bash` in the future. thanks!

lukeiannucci:
> got it, would be nice to offer `--port` options on `test-node.bash` in the future. thanks!

agreed :saluting_face:

alxiong:
I'm gonna squash merge this PR given how many iterations we've been through and none of them are important for the `main` history. 
I'd rather not pollute `main` badly.

lukeiannucci:
> I'm gonna squash merge this PR given how many iterations we've been through and none of them are important for the `main` history. I'd rather not pollute `main` badly.

thanks alex, thanks for all the hard work on this!

PR #471, opened 2025-08-26
try fix ci

lukeiannucci:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -195,7 +195,7 @@ jobs:
           --init &
         cd ..
     - name: Run test timeboost with nitro sequencer
-      run: just run_demo -s /tmp/stamp --ignore-stamp -c test-configs/c0 --rounds 10000 --yapper --nitro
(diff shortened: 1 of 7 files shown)


docker/timeboost.Dockerfile
```diff
@@ -1,10 +1,17 @@
...
 
 WORKDIR /app
 
 COPY . .
 RUN apt update && apt-get install -y protobuf-compiler libssl-dev
+RUN curl -L https://foundry.paradigm.xyz | bash && /root/.foundry/bin/foundryup
+ENV PATH="/root/.foundry/bin:${PATH}"
+RUN forge --version
+RUN rustup component add rustfmt --toolchain nightly
+RUN cargo install just
```
lukeiannucci, line 13: i dont think we need `just` i think maybe in contracts we could just run `cargo fmt`?

alxiong (approved):
Thanks for the help! all the changes looks great to me.

docker/timeboost.Dockerfile
alxiong: agree, fixed in [6b1dfbc](https://github.com/EspressoSystems/timeboost/pull/469/commits/6b1dfbc8ac0162dc87f586fa56b799b768e052aa)

PR #474, opened 2025-08-28
Use network name in Noise prologue.

twittner:
To avoid accidentally connecting peers of different networks we now include the network name (e.g. "sailfish") in the Noise prologue. The handshake will fail if differently named network peers try to connect.


lukeiannucci (approved):
This seems fine to me, but just curious how were different networks connecting to one another? dont we do some sort of port magic?

PR #476, opened 2025-08-29
log error if we cant connect to nitro node, delayed inbox cleanup

lukeiannucci:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/timeboost/src/forwarder/nitro_forwarder.rs
+++ b/timeboost/src/forwarder/nitro_forwarder.rs
@@ -8,10 +8,11 @@ use multisig::PublicKey;
 use sailfish::types::RoundNumber;
 use timeboost_proto::forward::forward_api_client::ForwardApiClient;
 use timeboost_proto::inclusion::InclusionList;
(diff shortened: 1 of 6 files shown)


timeboost/src/forwarder/nitro_forwarder.rs
```diff
@@ -44,7 +48,7 @@ impl NitroForwarder {
         round: RoundNumber,
         timestamp: Timestamp,
         txns: &[Transaction],
-        index: DelayedInboxIndex,
+        index: u64,
```
twittner, line 51: Should this not be `DelayedInboxIndex` and you turn it into a `u64` internally?

lukeiannucci: this feels like the ugliest way:
https://github.com/EspressoSystems/timeboost/pull/476/commits/b309aebc688270651c1f31854c6213512e3338be

perhaps i should make changes to the `DelayedInboxIndex` struct to make this better?

PR #488, opened 2025-09-05
Extend config file.

twittner:
Move several command-line arguments into config file.

### DIFF
--- a/timeboost-config/src/binaries/mkconfig.rs
+++ b/timeboost-config/src/binaries/mkconfig.rs
@@ -53,7 +53,7 @@ struct Args {
     #[clap(long, short)]
     stamp_dir: PathBuf,
 
-    /// Internal gPRC endpoints among nodes, default to same IP as sailfish with port + 3000
+    /// Internal gPRC endpoints among nodes, default to same IP as sailfish with port + 3
     #[clap(long)]
     internal_addr: Address,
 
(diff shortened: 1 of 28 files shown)


timeboost-config/src/binaries/mkconfig.rs
```diff
@@ -45,6 +45,14 @@ struct Args {
...
 
+    /// HTTP API of a timeboost node.
+    #[clap(long)]
+    http_api: Address,
+
+    /// Directory to store timeboost stamp file in.
+    #[clap(long, short)]
+    stamp_dir: PathBuf,
+
     /// Internal gPRC endpoints among nodes, default to same IP as sailfish with port + 3000
```
akonring, line 56:
```suggestion
    /// Internal gPRC endpoints among nodes, default to same IP as sailfish with port + 3
```

timeboost-config/src/node.rs
```diff
@@ -76,8 +89,11 @@ impl fmt::Display for NodeConfig {
...
 mod tests {
     const CONFIG: &str = r#"
+stamp = "/tmp/timeboost.0.stamp"
+
 [net.public]
 address = "127.0.0.1:8001"
+http_api = "127.0.0.1:8004"
 
 [net.internal]
 address = "127.0.0.1:11001"
```
akonring, line 99:
```suggestion
address = "127.0.0.1:8003"
```

akonring (approved):
👍

timeboost-config/src/node.rs
akonring: nvm. this is just a test.

PR #495, opened 2025-09-08
wip Contract event stream

alysiahuggins:
Closes #<ISSUE_NUMBER>


alxiong:
I merge `main` into this branch locally, and now i have to update the `timeboost-contract/build.rs` as I mentioned in the standup yesterday. 
I'll get it done in this PR instead of as a follow up.

cc @lukeiannucci

timeboost-contract/src/events.rs
```diff
@@ -0,0 +1,324 @@
...
+use alloy::{
+    primitives::Address,
+    providers::Provider,
+    rpc::types::{BlockNumberOrTag, Filter},
+    sol_types::SolEvent,
+};
+use anyhow::Result;
+
+pub struct KeyManagerEventMonitor<P> {
+    provider: P,
```
alxiong, line 12: I wonder if we should hold `Arc<P>` instead, given that we are cloning it to create this struct and to create `CommitteeManager`.

timeboost-contract/src/committee.rs
```diff
@@ -0,0 +1,234 @@
...
+            provider,
+            contract_addr,
+        }
+    }
+
+    pub async fn get_committees_for_startup(
+        &self,
+        current_id: u64,
+        previous_id: Option<u64>,
+    ) -> Result<(KeyManager::Committee, Option<KeyManager::Committee>)> {
```
alxiong, line 23: i find `struct CommitteeManager` an overkill, and i also find this function a bit unnecessary. 

usually wherever i need to fetch committee info, i simply have two lines, one line to get a `KeyManager` contract instance, another line to `.getCommitteeById()` on it. that's it. 

if we use this struct, we first need to `CommitteManager::new()`, then call this `for_startup()`. 
imo, the former is clearer and less indirection.


The decision to fetch "which committee" during startup could be logic that resides in `timeboost` code, not here. 
I vote for dropping this file.

timeboost/src/event_monitor.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub fn new(provider: P, contract_addr: Address, config: EventMonitoringConfig) -> Self {
+        Self {
+            provider,
+            contract_addr,
+            config,
+        }
+    }
+
+    /// Start monitoring contract events in the background
+    pub async fn start_monitoring(self) -> Result<()> {
```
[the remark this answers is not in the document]
alxiong: btw, `alloy_pubsub`,  which I never use before, also provides a full-fledged pubsub layer, we can automate the reconnection, duplication handling, reorder etc. 

```rust
use alloy_primitives::{Address, B256, U256};
use alloy_pubsub::{connect::PubSubConnect, service::PubSubService};
use alloy_rpc_types::{Filter, Log};
use alloy_sol_types::sol;
use alloy_transport_ws::WsConnect;
use serde_json::Value;
use tokio::task;

sol! {
    event Transfer(address indexed from, address indexed to, uint256 value);
}

#[tokio::main]
async fn main() -> eyre::Result<()> {
    // Example ERC20 (USDC mainnet)
    let contract: Address = "A0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48".parse()?;

    // Topic0: keccak("Transfer(address,address,uint256)")
    let transfer_sig: B256 = alloy_primitives::keccak256("Transfer(address,address,uint256)");

    // JSON-RPC filter: only this contract + this event signature
    let filter = Filter::new().address(contract).topic0(transfer_sig);

    // 1. WebSocket provider
    let ws = WsConnect::new("wss://mainnet.infura.io/ws/v3/YOUR_PROJECT_ID");

    // 2. Setup pubsub frontend + service
    let (frontend, service) = PubSubService::new();

    // 3. Spawn service in background
    task::spawn(async move {
        if let Err(err) = PubSubConnect::new(ws).serve(service).await {
            eprintln!("pubsub service ended: {err:?}");
        }
    });

    // 4. Subscribe to logs
    let mut sub = frontend
        .subscribe("eth_subscribe", ["logs", serde_json::to_value(filter)?])
        .await?;

    // 5. Process messages
    while let Some(msg) = sub.recv().await {
        // Parse into a Log type
        let val: Value = msg;
        if let Ok(log) = serde_json::from_value::<Log>(val.clone()) {
            // Decode with the Transfer ABI
            if let Ok(decoded) = Transfer::decode_log(&log, true) {
                println!(
                    "Transfer: from={} to={} value={}",
                    decoded.from,
                    decoded.to,
                    decoded.value
                );
            } else {
                eprintln!("Failed to decode log: {val:?}");
            }
        } else {
            eprintln!("Unexpected message: {val:?}");
        }
    }

    Ok(())
}
```

What do we prefer? a simple but manual event stream: https://alloy.rs/examples/subscriptions/event_multiplexer 

or using pubsub?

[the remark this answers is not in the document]
lukeiannucci: I think this is a good idea if we can refactor it to use a websocket. i didnt know alloy had this support. it is also worth noting committee changes will be very infrequent per my understanding, so im just wondering if its possible the websocket will constantly be disconnecting and reconnecting, which is still fine. perhaps i can also refactor the delayed inbox task to also use a websocket connection.

timeboost-contract/src/events.rs
alysiahuggins: thanks for this,updated here https://github.com/EspressoSystems/timeboost/pull/495/commits/7a586666a5b7ba278cfb7efcc10553ffff0d9168

alxiong: thanks, btw the variable name doesn't need `_arc`. we don't usually expose the kind of smart pointer the data resides it.

alxiong:
closing this now, superseded by #502

PR #513, opened 2025-09-16
fix: prevent block loss on transaction size overflow in submitter

Fibonacci747:
This change fixes a bug where a block popped from outbox could be dropped when the assembled transaction reached the size limit. We now push the overflowing block back to the front of outbox, allow packing exactly up to the limit, and warn-drop blocks that individually exceed the configured max size. These changes are necessary to prevent silent loss of blocks, avoid deadlocks caused by unshippable blocks, and ensure consistent, deterministic transaction assembly behavior.

### DIFF
--- a/timeboost-builder/src/submit.rs
+++ b/timeboost-builder/src/submit.rs
@@ -181,10 +181,18 @@ impl Sender {
                         continue;
                     }
                     let n = minicbor::len(&b);
-                    if size + n < self.size_limit {
+                    // If a single block exceeds the transaction size limit, skip it with a warning
+                    // to avoid permanently blocking the queue with an unshippable item.
+                    if n > self.size_limit {
+                        warn!(node = %self.label, block = %b.cert().data().num(), size = %n, limit = %self.size_limit, "block exceeds max transaction size; dropping");
+                        continue;
+                    }
+                    if size + n <= self.size_limit {
                         size += n;
                         transaction.push(b)
                     } else {
+                        // Re-queue the block to avoid dropping it when current transaction is full.
+                        outbox.push_front(b);
                         break;
                     }
                 }


timeboost-builder/src/submit.rs
```diff
@@ -181,10 +181,18 @@ impl Sender {
...
+                    if n > self.size_limit {
+                        warn!(node = %self.label, block = %b.cert().data().num(), size = %n, limit = %self.size_limit, "block exceeds max transaction size; dropping");
+                        continue;
+                    }
+                    if size + n <= self.size_limit {
                         size += n;
                         transaction.push(b)
                     } else {
+                        // Re-queue the block to avoid dropping it when current transaction is full.
+                        outbox.push_front(b);
```
twittner, line 195:
```suggestion
                        outbox.push_front(b);
```
Good catch!

timeboost-builder/src/submit.rs
```diff
@@ -181,10 +181,18 @@ impl Sender {
                         continue;
                     }
                     let n = minicbor::len(&b);
-                    if size + n < self.size_limit {
+                    // If a single block exceeds the transaction size limit, skip it with a warning
+                    // to avoid permanently blocking the queue with an unshippable item.
+                    if n > self.size_limit {
+                        warn!(node = %self.label, block = %b.cert().data().num(), size = %n, limit = %self.size_limit, "block exceeds max transaction size; dropping");
+                        continue;
+                    }
```
twittner, line 189:
```suggestion
                    if n > self.size_limit {
                        error!(
                            node  = %self.label,
                            block = %b.cert().data().num(),
                            size  = %n,
                            limit = %self.size_limit,
                            "block exceeds max transaction size; dropping"
                        );
                        continue;
                    }
```
It is not currently reflected in the code but a single block will never exceed that max. transaction size. That being said, logging an error here is fine.

twittner (approved):

PR #544, opened 2025-09-30
WIP AWS setup in docker

akonring:


nitro-config/sequencer_config.json
```diff
@@ -0,0 +1,137 @@
...
+    },
+    "ws": {
+        "addr": "0.0.0.0"
+    },
+    "http": {
+        "addr": "0.0.0.0",
+        "vhosts": "*",
+        "corsdomain": "*"
+    },
+    "log-level": "DEBUG"
```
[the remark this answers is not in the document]
akonring: Yes, fixed: [ff5c071](https://github.com/EspressoSystems/timeboost/pull/544/commits/ff5c071fd7aeaea454d9bb3028ae3d3127faf466)

test-configs/docker/committee.toml
```diff
@@ -1,42 +1,42 @@
 id = 0
```
lukeiannucci, line 1: do we still need docker config folder?

timeboost-config/src/binaries/mkconfig.rs
```diff
@@ -248,8 +255,9 @@ impl Args {
         Ok(())
     }
 
-    fn adjust_addr(&self, i: u8, base: &Address) -> Result<Address> {
-        match self.mode {
+    fn adjust_addr(mode: Mode, i: u8, base: &Address) -> Result<Address> {
```
lukeiannucci, line 258: just curios why do we get rid of `self`?

docker-compose.block-maker.yml
```diff
@@ -0,0 +1,309 @@
+services:
```
lukeiannucci, line 1: this is awesome 👍

lukeiannucci (approved):
this is absolutely awesome, thanks for the hard work on this!!

timeboost-config/src/binaries/mkconfig.rs
akonring: Right, `Args` has both nitro-mode and public-mode now so it is not clear which mode that is subject to adjustment. Moved `adjust_addr` so it is now a method on `Mode` which seems more natural.

PR #545, opened 2025-10-01
Add support for running scenarios.

twittner:
`run-committee.rs` optionally accepts a scenario TOML file which specifies steps that should be taken. A step consists of an action and an optional delay. Currently the only supported actions are:

- remove a file
- start a node
- stop a node
- exit

The file `scenario.rs` implements the scenario configuration.

### DIFF
--- a/justfile
+++ b/justfile
@@ -3,17 +3,8 @@ export RUSTDOCFLAGS := '-D warnings'
 log_levels  := "RUST_LOG=timeboost=debug,sailfish=debug,cliquenet=debug,tests=debug"
 run_as_root := if env("CI", "") == "true" { "sudo" } else { "run0" }
 
-####################
-###BUILD COMMANDS###
-####################
-
 build *ARGS:
   cargo build {{ARGS}}
-
-update-submodules:
-  git submodule update --remote --recursive
-  cd timeboost-proto && cargo build
-  cd ../contracts && forge build
 
 build_release *ARGS:
   cargo build --release --workspace --all-targets {{ARGS}}
@@ -35,9 +26,6 @@ build-port-alloc:
 build-test-utils:
   cargo build --release -p test-utils --all-features
 
-####################
-###CHECK COMMANDS###
-####################
 clippy:
   cargo clippy --workspace --lib --tests --benches -- -D warnings
 
@@ -58,19 +46,11 @@ fmt_check:
 
 lint: clippy fmt_check
 
-fix:
-  cargo fix --allow-dirty --allow-staged
-
 ci_local:
   just build && just lint && just test_ci --release && \
   just run_demo --ignore-stamp --yapper -c test-configs/c0 && \
   just run-sailfish-demo && just build_docker
 
-bacon: clippy check fmt
-
-####################
-####RUN COMMANDS####
-####################
 run_integration: build_docker
   -docker network create --subnet=172.20.0.0/16 timeboost
   docker compose -f docker-compose.yml -f docker-compose.metrics.yml up -d
@@ -162,9 +142,6 @@ mkconfig_nitro DATETIME *ARGS:
 verify_blocks *ARGS:
   cargo run --release --bin block-verifier {{ARGS}}
 
-####################
-####TEST COMMANDS###
-####################
 test *ARGS: build-port-alloc
   target/release/run --spawn target/release/port-alloc cargo nextest run -- {{ARGS}}
   @if [ "{{ARGS}}" == "" ]; then cargo test --doc; fi
@@ -190,7 +167,7 @@ test-contract-deploy *ARGS:
   scripts/test-contract-deploy {{ARGS}}
 
 test-all: build_release build-test-utils
-  env RUST_LOG=timeboost_builder::submit=debug,block_checker=info,warn \
+  env RUST_LOG=timeboost_builder::submit=debug,block_checker=info,error \
   target/release/run \
     --verbose \
     --timeout 120 \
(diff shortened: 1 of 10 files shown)


test-utils/src/binaries/run-committee.rs
```diff
@@ -132,15 +147,83 @@ async fn main() -> Result<()> {
...
+        for (node, cmd) in commands {
+            if !subset.contains(&node) {
+                if args.verbose {
+                    eprintln!("ignoring node {node} command");
+                }
+                continue;
+            }
+            tasks.spawn(async move {
+                if args.verbose {
+                    eprintln!("spawing timeboost: \"{cmd}\"");
```
akonring: Thought: Should the verbosity vary depending on if a scenario is provided or not?

test-utils/src/scenario.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub action: Action,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+#[serde(rename_all = "kebab-case", tag = "kind")]
+pub enum Action {
+    Remove { files: Vec<PathBuf> },
+    StartNode { node: String, label: Option<String> },
+    StopNode { node: String, label: Option<String> },
+    Exit,
```
akonring: Thought: What is Exit action supposed to support (if not scenario exit)?

timeboost-sequencer/src/include.rs
akonring:
```suggestion
            if n >= self.committee.one_honest_threshold().get() {
```

test-utils/src/binaries/run-committee.rs
akonring:
```suggestion
                    eprintln!("spawning timeboost: \"{cmd}\"");
```

akonring (approved):
Nice — this is an excellent addition to the testing suite.

**Scenario**
As I mentioned, I’m still not entirely clear on the intended use of the Exit action and the `default.toml` scenario. The way configuration files are organized and applied for network testing and scenarios also feels a bit unintuitive.

**Includer Cache**
It’s great that we’re now applying the cache before adding items to the bundle queue. That also aligns well with the [recommendation in the spec](https://github.com/OffchainLabs/decentralized-timeboost-spec/blob/e314d70f9d0ebabccea17a59804e3b90ec96b936/inclusion.md?plain=1#L123).

However, I don’t think this resolves [the core issue](https://app.asana.com/1/1208976916964769/task/1211805884648915/comment/1211805884648920?focus=true) of inconsistent caches across nodes. Even if honest nodes insert entries immediately, some duplicates may already have entered the queue before the cache can filter them out.

As I understand it, the root cause of the cache inconsistency is that insertion depends on existing cache entries. Thus, a recovering node (with an empty cache) can never be guaranteed to have a consistent cache merely by observing (8 rounds of) output of consensus.

A potential solution might be to make cache insertion independent of the existing cache state. One approach could be to add the same entry multiple times (once for each round it appears more than t + 1). This seems consistent with [the requirement in the specification](https://github.com/OffchainLabs/decentralized-timeboost-spec/blob/e314d70f9d0ebabccea17a59804e3b90ec96b936/inclusion.md?plain=1#L143).

**Duplicate Transactions**
After some debugging, it looks like Yapper samples the same transaction every time when using the current Arbitrary setup for Legacy transactions. While we should fix that, it’s actually useful that it exposed this duplicate-related issue — which we should be resilient to.

justfile
```diff
@@ -255,7 +244,11 @@ test-dyn-comm: build-release-until build-test-utils
     --spawn "1:anvil --port 8545" \
     --run   "2:sleep 2" \
     --run   "3:scripts/deploy-contract -c test-configs/c0/committee.toml -u http://localhost:8545" \
-    --spawn "4:target/release/run-committee -c test-configs/c0/ --max-nodes 5 --until 2000" \
+    --spawn "4:target/release/run-committee \
+        -c test-configs/c0/ \
+        --ignore-stamp \
```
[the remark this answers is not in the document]
twittner: I would not want to make scenarios mandatory. It is more work to write one and one can run meaningful tests without one.

akonring:
> Can you elaborate on this? How would you want configuration to be organised?

From a high-level perspective, `net` and `scenario` seem to represent independent “layers” that can be applied on top of an existing test/committee configuration. For `net`, everything (including node configs) is placed under `linux`, whereas for `scenario` we take the scenario config from the `scenario` folder but node configs from elsewhere — though the keys must still align with what’s used in the scenario.
I don’t think this needs to be addressed here, just sharing some thoughts while trying to navigate this.

> I think consistency can be achieved if we can ensure that a bundle only appears in more than t candidate lists when not already in the includer caches. As you pointed out, we currently allow duplicates in our transaction queue and (independent of this) I think we should only allow a bundle to exist once in the queue.

Does this actually fix the underlying issue? Sailfish might consume a bundle when proposing a candidate list for round *r*, after which an identical bundle could be added to the queue again before consensus is reached (and before the cache is updated) for that same round.

> The question is though if members add the bundle to their inclusion list or not? If they do, it would make the cache redundant as members always add bundles which appear in at least t + 1 candidate lists.

Both approaches should work I think. If they do, we would deviate from the recommended reference strategy, but we’d still produce consistent inclusion lists. The cache is not redundant if we apply it to the bundle queue.

> If they don't, how can the recovering node distinguish this from the case where the bundle should be added?

I might be misunderstanding, but the bundle should always be added by all nodes (recovering or healthy) to the cache for that specific round. Healthy nodes would already have included the bundle in their inclusion list depending on the cache state prior to that insertion.

> AFAIU this is only the reference implementation strategy and in fact the whole inclusion cache is merely an optimisation which is not required.

Right — we could choose not to follow the reference recommendation. The key point is that inclusion lists for honest (recovered) nodes remain consistent.

akonring:
> No, Sailfish does not consume a bundle when proposing. It stays in the queue and could not be added again before consensus. 

Right, that makes sense. Then deduplication should be sufficient.

akonring:
> Yes, I was wondering though, what should be done differently?

Would sample and co-locate node/committee configs in the scenarios folder to be consistent with the linux/net approach (YMMV).

